PR #5, opened 2024-04-09
Write Unit tests

AryanGodara:
In progress


crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -179,4 +178,36 @@ mod tests {
...
+        use std::fs::{read_to_string, File};
+        use std::io::{self, BufRead};
+        use serde_json::Error;
+
+        #[test]
+        fn state_update_to_blob_data_works() {
+            let state_update_path = "test-utils/stateUpdate.json".to_owned();
+            let contents = read_to_string(state_update_path).expect("Couldn't find or load that file.");
+
+            let v: Result<StateUpdate, Error> = serde_json::from_str(&contents.as_str());
```
apoorvsadana, line 193: you can use .expect here with the error message directly instead of doing a match later on. .expect panics if there's an Err

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -179,4 +178,36 @@ mod tests {
...
+
+    mod test_state_update_to_blob_data {
+        use super::*;
+        use std::fs::{read_to_string, File};
+        use std::io::{self, BufRead};
+        use serde_json::Error;
+
+        #[test]
+        fn state_update_to_blob_data_works() {
+            let state_update_path = "test-utils/stateUpdate.json".to_owned();
```
apoorvsadana, line 190: is this file uploaded? i don't find it

crates/orchestrator/tests/common/constants.rs
```diff
@@ -0,0 +1,3 @@
+pub const MADARA_RPC_URL: &str = "http://localhost:9944";
+pub const MONGODB_CONNECTION_STRING: &str = "mongodb+srv://admin:Km6MuqgzRcRpSH2@cluster0-1.hhmuw56.mongodb.net/";
```
apoorvsadana, line 2: you might want to revoke keys if you committed this by mistake :)

crates/orchestrator/tests/config/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+    #[case] rpc_url: String, #[case] db_url: String, #[case] da_url: String 
+) {
+    let config = get_or_init_config(rpc_url, db_url, da_url).await;
+    config.starknet_client();
+    config.da_client();
+    config.database();
+    config.queue();
+}
+
+#[ignore = "Run this separately as it will fail the other tests (config can't be created with each test)"]
```
apoorvsadana, line 58: why can't the config be created with each test case?

crates/orchestrator/tests/config/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+    #[future] get_or_init_config: &'static dyn Config,
+) {
+    let config = get_or_init_config.await;
+    // let result = config.starknet_client().get_state_update(BlockId::Number(0u64)).await;
+    let result = config.starknet_client().block_number().await;
+    assert!(result.is_ok(), "Failed to run starknet_client()");
+}
+
+#[rstest]
+#[should_panic]
```
apoorvsadana, line 80: if we can use the custom init_config method for test cases as mentioned above, we can check if the mock was called over here

crates/orchestrator/tests/config/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+
+#[rstest]
+#[tokio::test]
+async fn test_config_database(
+    #[future] get_or_init_config: &'static dyn Config,
+    default_job_item: JobItem,
+) {
+    let config = get_or_init_config.await;
+    let job = default_job_item;
+    let result = config.database().create_job(job).await;
```
apoorvsadana, line 100: same as above

crates/orchestrator/tests/jobs/da_job/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    let v: Result<StateUpdate, Error> = serde_json::from_str(&contents.as_str());
+
+    let state_update: StateUpdate = match v {
+        Ok(state_update) => state_update,
+        Err(e) => panic!("Couldn't parse the JSON file: {}", e),
+    };
+
+    // let blob_data = state_update_to_blob_data(630872, state_update);
+    state_update
```
apoorvsadana, line 40: same comments as previous state update code

crates/orchestrator/tests/jobs/da_job/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+) {
+    let config = config.await;
+    let job = DaJob.create_job(config, String::from("0")).await;
+    assert!(job.is_ok());
+
+    let job = job.unwrap();
+    
+    let job_type = job.job_type;
+    assert_eq!(job_type, JobType::DataSubmission);
+    assert_eq!(job.metadata.values().len(), 0, "metadata should be empty");
```
apoorvsadana, line 61: not a valid test point, maybe other job types will add meta data. better to just test the 

1. type is DataSubmission
2. check if job_id is present
3. status is created
4. version is 0

crates/orchestrator/tests/jobs/da_job/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+#[rstest]
+// #[should_panic]
+#[tokio::test]
+async fn test_verify_job(
+    #[future] #[from(get_or_init_config)] config: &'static dyn Config,
+    #[from(default_job_item)] job_item: JobItem,
+) {
+    let config = config.await;
+
+    assert!(DaJob.verify_job(config, &job_item).await.is_err());
```
apoorvsadana, line 74: can check if the da handler's `verify_inclusion` was called

crates/orchestrator/tests/jobs/da_job/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+
+#[rstest]
+fn test_max_verification_attempts() {
+    assert_eq!(DaJob.max_verification_attempts(), 3);
+}
+
+#[rstest]
+fn test_verification_polling_delay_seconds() {
+    assert_eq!(DaJob.verification_polling_delay_seconds(), 60);
+}
```
apoorvsadana, line 137: we can skip these tests. these values will be configurable, we don't want to change the test everytime we change the value

crates/orchestrator/tests/config/mod.rs
AryanGodara: The config is created only once because of this ( `pub static CONFIG: OnceCell<AppConfig> = OnceCell::const_new();`)

And in this test, I'm trying to create an invalid config.
But since a common config is already created for the other tests, no new config gets created for this particular test, and so, this behaviour can't be tested alongside all other tests, and has to be run separately

crates/orchestrator/tests/common/constants.rs
AryanGodara: This is a free mongo cluster. I thought it'd be simpler to store it here, plus I can just close it down if it's overused 😅 .
Feel free to use this for tests, as it has already set up the DBs ig

PR #7, opened 2024-05-24
Publish State Diff to Eth

Mohiiit:
functionality added to send state diff data to eth as blob
added tests

## Summary by CodeRabbit

- **New Features**
  - Introduced new functions for data conversion, FFT transformation, and blob creation.
  - Added static variables for BLS12-381 modulus and related constants.

- **Bug Fixes**
  - Updated functions to handle Ethereum blob size limits per transaction.

- **Improvements**
  - Refactored `state_update_to_blob_data` for better nonce retrieval and data processing.
  - Enhanced `publish_state_diff` method to support new transaction handling logic.

- **Documentation**
  - Updated environment variable requirements for Ethereum RPC URL and private key.

- **Tests**
  - Added test cases for new functions and data handling logic.


crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -65,11 +120,93 @@ impl Job for DaJob {
...
+}
+
+async fn state_update_to_blob_data(
+    block_no: u64,
+    state_update: StateUpdate,
+    config: &Config,
+) -> Result<Vec<FieldElement>> {
     let state_diff = state_update.state_diff;
     let mut blob_data: Vec<FieldElement> = vec![
         // TODO: confirm first three fields
```
apoorvsadana, line 206:
```suggestion
```

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -134,40 +264,61 @@ fn state_update_to_blob_data(block_no: u64, state_update: StateUpdate) -> Vec<Fi
...
+    let binary_representation = format!("{:b}", num_changes);
+    let padded_binary_string = format!("{:0>64}", binary_representation);
+    binary_string += &padded_binary_string;
+
+    let biguint = BigUint::from_str_radix(binary_string.as_str(), 2).expect("Invalid binary string");
+
+    // Now convert the BigUint to a decimal string
+    let decimal_string = biguint.to_str_radix(10);
+
+    let word = FieldElement::from_dec_str(&decimal_string).expect("issue while converting to fieldElement");
```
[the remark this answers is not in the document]
apoorvsadana: does this work? looks cleaner

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -180,36 +348,79 @@ mod tests {
...
+        let mut file = File::create(file_path).expect("not able to create the files");
+        for number in numbers {
+            writeln!(file, "{}", number).expect("unable to write to the file");
+        }
+    }
+
+    #[rstest]
+    // #[case(631861, "src/jobs/test_utils/test_blob_631861.txt")]
+    // #[case(638353, "src/jobs/test_utils/test_blob_638353.txt")]
+    // #[case(639404, "src/jobs/test_utils/test_blob_639404.txt")]
```
apoorvsadana, line 361: why comment out?

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -180,36 +348,79 @@ mod tests {
...
+    #[case(640641, "src/jobs/test_utils/test_blob_640641.txt")]
+    #[tokio::test]
+    async fn test_state_update_to_blob_data(#[case] block_no: u64, #[case] file_path: &str) {
+        // @note: not yet done
+        let config = init_config(
+            Some(format!("https://starknet-mainnet.infura.io/v3/3753abe17c124d088f4e68d58f257452")),
```
apoorvsadana, line 367: let's store the state update json locally. not a good practice to rely on external API calls for unit tests

crates/da_clients/ethereum/test_utils/vec_8_block_630872.txt
apoorvsadana, line 1: where's this used?

crates/orchestrator/src/jobs/test_utils/test_blob_631861.txt
apoorvsadana, line 1: these files should be inside the da_job folder. also, we should rename `test_utils` -> `test_data`, wdyt?

crates/orchestrator/src/jobs/da_job/mod.rs
Mohiiit: how will fetch the nonce for the addresses where state_diff is non-zero? using mocks?

apoorvsadana: ya, we would need to mock the rpc endpoint like we do in some test currently using httpmock

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -65,11 +120,93 @@ impl Job for DaJob {
     }
 }
 
-fn state_update_to_blob_data(block_no: u64, state_update: StateUpdate) -> Vec<FieldElement> {
+fn fft_transformation(elements: Vec<BigUint>) -> Vec<BigUint> {
+    // @note: this should be in constants ideally
```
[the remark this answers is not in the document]
Mohiiit: you mean move constants here?

crates/orchestrator/src/jobs/da_job/mod.rs
Mohiiit: should I store the nonce values for the test case?

crates/da_clients/ethereum/src/lib.rs
```diff
@@ -1,38 +1,199 @@
...
+            prepare_sidecar(&_state_diff, &trusted_setup).await?;
+        let sidecar = BlobTransactionSidecar::new(sidecar_blobs, sidecar_commitments, sidecar_proofs);
+
+        // chain id should be an env variable or should be in config?
+        let tx = TxEip4844 {
+            chain_id: 17000, // Holesky 17000 sepolia 11155111
+            nonce: 1,
+            gas_limit: 30_000_000,
+            max_fee_per_gas: 10000000100, //estimation.max_fee_per_gas.to_string().parse()?,
+            max_priority_fee_per_gas: 200000010,
```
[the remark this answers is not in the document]
Mohiiit: it was not there when I checked last, checking again

crates/orchestrator/src/jobs/da_job/test_data/nonces_from_block_640641.txt
```diff
@@ -0,0 +1,298 @@
...
+    "nonce": "0",
+    "address": "3488796095165551003812490975976509991633322013578458084167226256699277019952"
+  },
+  {
+    "nonce": "0",
+    "address": "773419650418337517107856101029931836850298774485466530128029681791734852785"
+  },
+  {
+    "nonce": "a",
+    "address": "239662865362034240643029484020826104800516253213877576945698001454992152761"
```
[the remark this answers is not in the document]
Mohiiit: nonce value is in hex, it's correct

crates/da_clients/ethereum/src/lib.rs
```diff
@@ -1,38 +1,212 @@
...
+
+        let eip1559_est = provider.estimate_eip1559_fees(None).await?;
+        let chain_id: u64 = provider.get_chain_id().await?.to_string().parse()?;
+
+        let max_fee_per_blob_gas: u128 = provider.get_blob_base_fee().await?.to_string().parse()?;
+        let max_priority_fee_per_gas: u128 = provider.get_max_priority_fee_per_gas().await?.to_string().parse()?;
+
+        let tx = TxEip4844 {
+            chain_id,
+            nonce: 0, // can be block number
```
unstark, line 55: If I'm not missing something, it would not be set right?
There should be NonceFiller in the alloy provider chain to handle that, couldn't find it explicitly set.

crates/da_clients/ethereum/src/lib.rs
```diff
@@ -1,38 +1,212 @@
...
+            gas_limit: 30_000_000,
+            max_fee_per_gas: eip1559_est.max_fee_per_gas.to_string().parse()?,
+            max_priority_fee_per_gas,
+            to: addr, // maybe to the L1 contract for verification??
```
unstark, line 59: to, input, and gas_limit should probably be function arguments, if the goal is to reuse Starknet core contract, that would be its address, updateStateKZG parameters, and appropriate gas_limit to account for the spendings

crates/da_clients/ethereum/src/lib.rs
apoorvsadana: ya makes sense, @Mohiiit let's fetch the nonce from the RPC itself

crates/da_clients/ethereum/src/lib.rs
apoorvsadana: right. i think we should fill in `to` and `gas_limit` for now. `input` can be done in a refactor later on when we integrate updateState

@Mohiiit we can ideally take `to` as an input and `gas_limit` can be estimated?

apoorvsadana: Hey @unstark, I have set the nonce by fetching it directly and used the `to` address as an input. Created an issue for the gas limit for now :)

PR #16, opened 2024-06-13
feat : added snos worker implementation and unit tests

ocdbytes:
- added SNOS worker implementation
- added SNOS worker tests

## Summary by CodeRabbit

- **New Features**
  - Introduced dynamic configuration management using `ArcSwap`.
  - Added database functions to retrieve the latest job by type and all jobs.
  - Included new `JobType::SnosRun` case in job handling.
  - Implemented a new test module for job creation and processing.

- **Enhancements**
  - Updated job functions to use reference configuration.
  - Improved worker-thread handling with corrected comments and terminology.

- **Dependencies**
  - Added `arc-swap` version 1.7.1 to the dependencies.

- **Tests**
  - Added extensive tests for job creation and worker interactions using mocks.


crates/orchestrator/src/config.rs
```diff
@@ -79,11 +80,28 @@ impl Config {
...
+pub async fn config() -> Guard<Arc<Config>> {
+    let cfg = CONFIG.get_or_init(|| async { ArcSwap::from_pointee(init_config().await) }).await;
+    cfg.load()
+}
+
+/// OnceCell only allows us to initialize the config once and that's how it should be on production.
+/// However, when running tests, we often want to reinitialize because we want to clear the DB and
+/// set it up again for reuse in new tests. By calling `config_force_init` we replace the already
+/// stored config inside `ArcSwap` with the new configuration and pool settings.
+pub async fn config_force_init(config: Config) {
```
apoorvsadana, line 98: let's only add this code if the test flag is enabled

crates/orchestrator/src/database/mod.rs
```diff
@@ -33,6 +34,8 @@ pub trait Database: Send + Sync {
     ) -> Result<()>;
 
     async fn update_metadata(&self, job: &JobItem, metadata: HashMap<String, String>) -> Result<()>;
+    async fn get_latest_job_by_type(&self, job_type: JobType) -> Result<Option<JobItem>>;
+    async fn get_all_jobs(&self, job_type: JobType) -> Result<Cursor<JobItem>>;
```
apoorvsadana, line 38: `Cursor` is a mongodb type and the DatabaseTrait is generic across the database you're using. So we should change this to `Vec` and handle the conversion for Cursor -> Vec inside the MongoDB function

crates/orchestrator/src/jobs/mod.rs
```diff
@@ -170,6 +170,7 @@
 fn get_job_handler(job_type: &JobType) -> Box<dyn Job> {
     match job_type {
         JobType::DataSubmission => Box::new(da_job::DaJob),
+        JobType::SnosRun => Box::new(da_job::DaJob),
```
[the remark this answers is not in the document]
apoorvsadana: @ocdbytes let's fix this

crates/orchestrator/src/tests/workers/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+use da_client_interface::MockDaClient;
+use httpmock::MockServer;
+use rstest::rstest;
+use serde_json::json;
+use std::collections::HashMap;
+use uuid::Uuid;
+
+#[rstest]
+#[tokio::test]
+async fn test_create_job() {
```
apoorvsadana, line 18:
```suggestion
async fn test_snos_worker() {
```

crates/orchestrator/src/tests/workers/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+            .returning(|_, _| Ok(None))
+            .call(&i.to_string(), &JobType::SnosRun)
+            .expect("Failed to call.");
+    }
+
+    // Creating jobs expectations
+    for i in 1..6 {
+        db.expect_create_job()
+            .returning(|_| Ok(get_job_item_mock_by_id("1".to_string())))
+            .call(get_job_item_mock_by_id(i.to_string()))
```
apoorvsadana, line 38: should we just call `get_job_item_mock_by_id` once and use it in both places?

crates/orchestrator/src/tests/workers/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+        db.expect_get_job_by_internal_id_and_type()
+            .returning(|_, _| Ok(None))
+            .call(&i.to_string(), &JobType::SnosRun)
+            .expect("Failed to call.");
+    }
+
+    // Creating jobs expectations
+    for i in 1..6 {
+        db.expect_create_job()
+            .returning(|_| Ok(get_job_item_mock_by_id("1".to_string())))
```
apoorvsadana, line 37: should 1 -> i

crates/orchestrator/src/tests/workers/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+    let mut db = MockDatabase::new();
+    let mut queue = MockQueueProvider::new();
+
+    // Mocking db functions
+    db.expect_get_latest_job_by_type().returning(|_| Ok(None)).call(JobType::SnosRun).expect("Failed to call.");
+    // Getting jobs for check expectations
+    for i in 1..6 {
+        db.expect_get_job_by_internal_id_and_type()
+            .returning(|_, _| Ok(None))
+            .call(&i.to_string(), &JobType::SnosRun)
```
apoorvsadana, line 30: we aren't supposed to `call` anywhere. instead, we need to tell the mock that this function should be called once (`times(1)`) and it should be called with args xyz (https://docs.rs/mockall/latest/mockall/#matching-arguments)

is this why you were saying that the test doesn't make sense because we're calling everything ourselves?

crates/orchestrator/src/tests/workers/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+#[rstest]
+#[tokio::test]
+async fn test_create_job() {
+    let server = MockServer::start();
+    let da_client = MockDaClient::new();
+    let mut db = MockDatabase::new();
+    let mut queue = MockQueueProvider::new();
+
+    // Mocking db functions
+    db.expect_get_latest_job_by_type().returning(|_| Ok(None)).call(JobType::SnosRun).expect("Failed to call.");
```
apoorvsadana, line 25: should we also do a test case for when the db returns a value? we could use `rstest` with `cases` so that we can use the same code only

crates/orchestrator/src/workers/snos.rs
```diff
@@ -8,7 +14,35 @@ impl Worker for SnosWorker {
     /// 1. Fetch the latest completed block from the Starknet chain
     /// 2. Fetch the last block that had a SNOS job run.
     /// 3. Create SNOS run jobs for all the remaining blocks
+    // TEST : added config temporarily to test
```
apoorvsadana, line 17: what's this?

crates/orchestrator/src/workers/snos.rs
```diff
@@ -8,7 +14,35 @@ impl Worker for SnosWorker {
...
     async fn run_worker(&self) {
-        todo!()
+        let config = config().await;
+        let provider = config.starknet_client();
+        let latest_block_number = provider.block_number().await.expect("Failed to fetch block number from rpc");
```
apoorvsadana, line 21: should we make workers return a Result and throw an error wherever we do `unwrap` or `expect`

crates/orchestrator/src/workers/snos.rs
```diff
@@ -8,7 +14,35 @@ impl Worker for SnosWorker {
...
+            return;
+        }
+
+        for x in latest_block_processed + 1..latest_block_number + 1 {
+            create_job(JobType::SnosRun, x.to_string(), HashMap::new())
+                .await
+                .expect("Error : failed to create job for snos workers.");
+        }
+
+        log::info!("jobs created !!");
```
apoorvsadana, line 46: trace log?

crates/orchestrator/src/tests/workers/mod.rs
ocdbytes: doesn't work tried this value move ho jaati h

apoorvsadana: try cloning?

crates/orchestrator/src/database/mod.rs
ocdbytes: There is no need for this function for now as test is working without it. Will remove this and if in future we need to add this function will surely follow the above method.

crates/orchestrator/src/tests/workers/mod.rs
ocdbytes: yup

crates/orchestrator/src/workers/snos.rs
ocdbytes: removed

crates/orchestrator/src/workers/snos.rs
ocdbytes: that should be the ideal case but ig error is handled in `create_job` function.

crates/orchestrator/src/workers/snos.rs
ocdbytes: removed. It is a redundant log

crates/orchestrator/src/tests/workers/mod.rs
ocdbytes: actually call behaviour is needed when test is run. 
So what happens is that if I don't add this call then test will throw an error stating that no expectation behaviour found for this particular call. Because this function is called in the code while creating the job to check whether the job already exists, If yes It will throw an error.

apoorvsadana: Hmm, the error sounds like the code didn't call this method but it should've been called. We don't use call for the da test. Can you double or share the exact error message?

crates/orchestrator/src/workers/snos.rs
apoorvsadana: but we're doing unwraps and expects in worker, this will crash the orchestrator

crates/orchestrator/src/tests/workers/mod.rs
ocdbytes: already tried

crates/orchestrator/src/database/mongodb/mod.rs
```diff
@@ -115,4 +115,23 @@ impl Database for MongoDb {
         self.update_job_optimistically(job, update).await?;
         Ok(())
     }
+
+    async fn get_latest_job_by_type(&self, job_type: JobType) -> Result<Option<JobItem>> {
+        let filter = doc! {
+            "job_type": mongodb::bson::to_bson(&job_type)?,
+        };
+        let find_options = FindOneOptions::builder().sort(doc! { "internal_id": -1 }).build();
```
unstark, line 123: AFAICS internal ID is string without additional restrictions https://github.com/unstark/madara-orchestrator/blob/3fe37331183cf3291aaef9d1e2664e54e86f0648/crates/orchestrator/src/jobs/types.rs#L104, so it's not necessarily a unique incrementing integer. We might need an additional field, e.g. created_at in microseconds to achieve what you want

crates/orchestrator/src/workers/snos.rs
unstark: +1 
In most cases it's better (imho) to propagate the error as far as possible and let the caller decide what to do with it rather than panic

crates/orchestrator/src/workers/snos.rs
```diff
@@ -8,7 +14,35 @@ impl Worker for SnosWorker {
     /// 1. Fetch the latest completed block from the Starknet chain
     /// 2. Fetch the last block that had a SNOS job run.
     /// 3. Create SNOS run jobs for all the remaining blocks
+    // TEST : added config temporarily to test
     async fn run_worker(&self) {
-        todo!()
+        let config = config().await;
+        let provider = config.starknet_client();
+        let latest_block_number = provider.block_number().await.expect("Failed to fetch block number from rpc");
+        let latest_block_processed_data = config
```
unstark, line 22: Could it be the case that there are gaps in the snos job executions?
For instance:

block 1: snos job succeeded
block 2: snos job failed
block 3: snos job succeeded
block 4: snos job pending

Ideally you should also search for failed ones and retry (if possible)

crates/orchestrator/src/config.rs
unstark: I wonder if it could work without the global config? Apart from the need to carry it through all methods, are there any reasons to use static initialization?

crates/orchestrator/src/database/mongodb/mod.rs
```diff
@@ -115,4 +115,23 @@ impl Database for MongoDb {
...
+            .await
+            .expect("Failed to fetch latest job by given job type"))
+    }
+
+    async fn get_all_jobs(&self, job_type: JobType) -> Result<Cursor<JobItem>> {
+        let filter = doc! {
+            "job_type": mongodb::bson::to_bson(&job_type)?,
+        };
+        Ok(self.get_job_collection().find(filter, None).await.expect("Failed to fetch jobs with given  job type"))
+    }
```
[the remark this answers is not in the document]
unstark: Given that internal ID is a string of variable length, definitely needs an index

crates/orchestrator/src/config.rs
apoorvsadana: Ya that's the only advantage. Otherwise we need to pass it everywhere although that does make testing easier. Do you think we should do the latter?

crates/orchestrator/src/tests/workers/mod.rs
apoorvsadana: Cloning a string should ideally work, what were you doing exactly?

crates/orchestrator/src/database/mongodb/mod.rs
apoorvsadana: So internal_id would actually be the block number but in String. The reason it was made a string is that jobs use this column to uniquely identify a job based on the type. Like, I want the SNOS run of block 6. Now jobs can have any sort of internal id (number, uuid etc.). So as a generic, a string param was used here. I am not sure if this is the best approach though but here, -1 should work because it's a number represented as string?

apoorvsadana: Ok nvm, this is incorrect. This function can be used by any job so if some job has a different form of internal id, this will break. However, the created_at might not be the best check either because what if we create jobs in parallel somewhere or after receiving something from a queue, there's no guarantee that the jobs would be created sequentially.

crates/orchestrator/src/workers/snos.rs
apoorvsadana: the job execution already has a retry logic. if something moved to a failed state, we have ideally already retried enough number of times and we should throw and alert and investigate manually, wdyt?

crates/orchestrator/src/database/mongodb/mod.rs
apoorvsadana: For now, we can explicitly rename this function to say `get_latest_job_by_internal_id` so it makes the purpose of the function clear. And then later, we can have another function of `get_latest_by_created_at` if some job is ok using that. wdyt?

crates/orchestrator/src/tests/workers/mod.rs
ocdbytes: resolved

crates/orchestrator/src/workers/snos.rs
ocdbytes: added error throwing in workers

crates/orchestrator/src/config.rs
unstark: To be fair, yes :) that would be much easier to test, also you'd be able to pass only necessary parts of the config instead of the entire struct

crates/orchestrator/src/database/mongodb/mod.rs
unstark: That would work (renaming), as for the created_at - even if there is a collision, the function would still do its job :)

crates/orchestrator/src/workers/snos.rs
unstark: Oh right, that makes sense 👍

PR #22, opened 2024-06-19
Feature: Implement proving worker

ocdbytes:
- Added proving worker
- Added proving worker test cases


crates/orchestrator/src/database/mongodb/mod.rs
```diff
@@ -127,4 +129,53 @@ impl Database for MongoDb {
             .await
             .expect("Failed to fetch latest job by given job type"))
     }
+
+    async fn get_successful_snos_jobs_without_proving(&self) -> Result<Vec<JobItem>> {
```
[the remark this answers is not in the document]
ocdbytes: yeah I will look into it and discuss and tell you. Can you explain a bit more on why we need pagination here in a db call ?

crates/orchestrator/src/jobs/proving_job/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+use crate::jobs::Job;
+use async_trait::async_trait;
+use color_eyre::Result;
+use std::collections::HashMap;
+use uuid::Uuid;
+
+pub struct ProvingJob;
+
+#[async_trait]
+impl Job for ProvingJob {
```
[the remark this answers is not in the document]
ocdbytes: sure will do it.

crates/orchestrator/src/database/mongodb/mod.rs
```diff
@@ -127,4 +129,53 @@ impl Database for MongoDb {
...
+                "$lookup": {
+                    "from": "jobs",
+                    "let": { "internal_id": "$internal_id" },
+                    "pipeline": [
+                        {
+                            "$match": {
+                                "$expr": {
+                                    "$and": [
+                                        { "$eq": ["$job_type", "ProofCreation"] },
+                                        { "$eq": ["$internal_id", "$$internal_id"] }
```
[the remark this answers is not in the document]
ocdbytes: Hey @unstark 

So I talked about this to apoorv and he told me that the interal_id field was just to make sure that for a particular job type there is a way to track the jobs internally so that's why a generic name `internal_id` was given to this field.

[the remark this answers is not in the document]
apoorvsadana: So @unstark my hypothesis was that the orchestrator can be designed in a way that we can also have jobs at the txn level later on. Maybe something like txn level proving or stream of txs to some XYZ place. So in that case, the internal_id will be the txn_hash. What are your thoughts on this? Does it feel like an overkill?

crates/orchestrator/src/database/mod.rs
```diff
@@ -36,6 +36,8 @@ pub trait Database: Send + Sync {
 
     async fn update_metadata(&self, job: &JobItem, metadata: HashMap<String, String>) -> Result<()>;
     async fn get_latest_job_by_type_and_internal_id(&self, job_type: JobType) -> Result<Option<JobItem>>;
+
+    async fn get_successful_snos_jobs_without_proving(&self) -> Result<Vec<JobItem>>;
```
apoorvsadana, line 40: I think the database trait shouldn't have functions with the name `snos` or `proving` as that seems to leak the internal types. We should generalise the function over all types and take the type as an input. Function can look like

```rust
fn get_jobs_without_successor(job_a_type, job_a_status, job_b_type, job_b_status)
```

wdyt?

crates/orchestrator/src/tests/workers/proving/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+        db.expect_get_successful_snos_jobs_without_proving()
+            .times(1)
+            .with()
+            .returning(move || Ok(get_job_item_mock_by_id_vec(5)));
+
+        prover_client.expect_submit_task().times(5).returning(|_| Ok("task_id".to_string()));
+    }
+
+    // Queue function call simulations
+    queue
```
apoorvsadana, line 68: should we make this times 5 or 4?

crates/orchestrator/src/database/mod.rs
ocdbytes: makes sense

crates/orchestrator/src/tests/workers/proving/mod.rs
ocdbytes: in first case it is 5 in second case when one job is unsuccessful that's why we are using 4 here.

crates/orchestrator/src/database/mongodb/mod.rs
```diff
@@ -133,16 +133,54 @@ impl Database for MongoDb {
...
+    ///
+    /// job_a_type : SnosRun
+    ///
+    /// job_a_status : Completed
+    ///
+    /// job_b_type : ProofCreation
+    ///
+    /// job_b_status : Status of Job B / None
+    ///
+    /// **IMP** : For now Job B status implementation is pending so we can pass None
```
apoorvsadana, line 158: let's not take job as a param for now because it allows others to call this function by mistake. and we can make the **IMP** a TODO then

Issue #24, opened 2024-06-24
feat: S3 implementation


apoorvsadana:
We need to add generalised `ObjectStorage` traits just like we do for Queue and Database at the moment. Then we also need add the implementation for S3 which can be used to store any blob of data on AWS S3.

PR #27, opened 2024-06-25
feat : added aws s3 adapter and data storage trait

ocdbytes:
## S3 Implementation

- Added data storage crate for getting and storing data on cloud
- Added AWS S3 trait and added .env values

Issue : #24


crates/orchestrator/src/data_storage/aws_s3/mod.rs
apoorvsadana, line 1: we're using `-` instead of `_` in file and crate names as a convention

crates/orchestrator/src/data_storage/aws_s3/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+use crate::data_storage::DataStorage;
+use async_trait::async_trait;
+use aws_sdk_s3::config::{Builder, Credentials, Region};
+use aws_sdk_s3::primitives::ByteStream;
+use aws_sdk_s3::{Client, Error};
+
+pub mod config;
+
+const SNOS_OUTPUT_FILE_NAME: &str = "snos_output.json";
+const KZG_FILE_NAME: &str = "kzg.txt";
```
apoorvsadana, line 12: s3 shouldn't be aware of this, these details can be in the snos job. also, do we need to store kzg now? i was thinking we can calculate the proof when we calculate state diffs in update state only

crates/orchestrator/src/data_storage/aws_s3/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+        // Building AWS S3 config
+        let client = Client::from_conf(conf);
+
+        Self { client, config }
+    }
+}
+
+#[async_trait]
+impl DataStorage for AWSS3 {
+    async fn get_snos_data_for_block(&self, block_number: u128) -> Result<StarknetOsOutput, Error> {
```
apoorvsadana, line 45: same as above, S3 should have no idea on what snos is. s3 should only have get and put functions.

.env.example
```diff
@@ -13,4 +13,10 @@ MONGODB_CONNECTION_STRING
 
 # SQS
 AWS_ACCESS_KEY_ID=
-AWS_SECRET_ACCESS_KEY=
\ No newline at end of file
+AWS_SECRET_ACCESS_KEY=
+
+# S3
+AWS_S3_KEY_ID=
+AWS_S3_KEY_SECRET=
```
apoorvsadana, line 20: so, I think, as long as `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are in the env variables, aws sdk automatically fetches the config. like, we never initialise the SQS config in the orchestrator, it just works. although we do use a wrapper library as well so maybe that handles it automatically but i mostly think its the aws sdk itself. and in that case, i think we can just have one aws access key and secret for all AWS services

crates/orchestrator/src/data_storage/aws_s3/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+        // Building AWS S3 config
+        let client = Client::from_conf(conf);
+
+        Self { client, config }
+    }
+}
+
+#[async_trait]
+impl DataStorage for AWSS3 {
+    async fn get_data_for_block(&self, key: &str) -> Result<StarknetOsOutput, Error> {
```
apoorvsadana, line 42: it should be just `get_data` and it should return just bytes. if we hardcode `StarknetOsOutput`, I can't use S3 later on with the DA layer for example in case I want to store it in s3

crates/orchestrator/src/data_storage/aws_s3/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+            .await
+            .expect("Failed to convert body into AggregatedBytes.");
+        let data_bytes = data_stream.into_bytes();
+        let json: StarknetOsOutput =
+            serde_json::from_slice(&data_bytes).expect("Failed to convert data_bytes into JSON.");
+
+        Ok(json)
+    }
+
+    async fn put_data_for_block(&self, data: StarknetOsOutput, key: &str) -> Result<usize, Error> {
```
apoorvsadana, line 62: same as above

crates/orchestrator/src/data_storage/aws_s3/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+//     use rstest::rstest;
+//     use crate::config::config;
+//     use crate::data_storage::aws_s3::AWSS3;
+//     use crate::data_storage::aws_s3::config::AWSS3Config;
+//     use crate::data_storage::{DataStorage, DataStorageConfig};
+//     use crate::data_storage::types::StarknetOsOutput;
+//
+//     #[rstest]
+//     #[tokio::test]
+//     async fn test_put_data_in_bucket() {
```
apoorvsadana, line 90: is this pending?

PR #34, opened 2024-07-09
feat : kzg proof calculation during state update & state update transaction 

ocdbytes:
- Adding a function to calculate the kzg proof of x_0 locally.
- Adding program output vars into the transaction to update the state.
- Adding tests for updating the state.
- Added the function to update the state and publish blob on ethereum in state update job.
- Removed all the fetch_from_test argument.
- Storage Client mocking added in tests.
- Added build proof function to settlement client and added enum for inputs and outputs for the function.


crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -272,9 +280,31 @@ async fn state_update_to_blob_data(
         blob_data.push(*compiled_class_hash);
     }
 
+    // saving the blob data of the block to S3 bucket (if not running in test mod)
+    if !test_mode {
```
apoorvsadana, line 284: test_mode shouldn't be a flag of a function that's used normally. ideally we can mock the S3 client so that when it's called nothing happens?

crates/orchestrator/src/jobs/state_update_job/kzg.rs
```diff
@@ -0,0 +1,124 @@
...
+use crate::data_storage::aws_s3::AWSS3;
+use crate::data_storage::{DataStorage, DataStorageConfig};
+use crate::jobs::state_update_job::CURRENT_PATH;
+use alloy::eips::eip4844::BYTES_PER_BLOB;
+use c_kzg::{Blob, Bytes32, KzgCommitment, KzgProof, KzgSettings};
+
+/// Build KZG proof for a given block
+pub async fn build_kzg_proof(
+    block_number: u64,
+    fetch_from_tests: Option<bool>,
```
apoorvsadana, line 11: same as above, `fetch_from_tests` shouldn't be an argument. normal code should almost never change for test cases. during the test case, we should try to mock functions to get the functionality we're looking for. In this specific case, I think mocking `s3_client.get_data` will do the job?

crates/orchestrator/src/jobs/state_update_job/kzg.rs
```diff
@@ -0,0 +1,124 @@
...
+) -> color_eyre::Result<(Vec<u8>, KzgProof)> {
+    let blob_data = fetch_blob_data_for_block(block_number, fetch_from_tests).await?;
+    let mut fixed_size_blob: [u8; BYTES_PER_BLOB] = [0; BYTES_PER_BLOB];
+    fixed_size_blob.copy_from_slice(blob_data.as_slice());
```
apoorvsadana, line 15: `copy_from_slice` can panic, we should instead do

```rust
let fixed_size_blob: [u8; BYTES_PER_BLOB] = blob_data.as_slice().try_into()?;
```

crates/orchestrator/src/jobs/state_update_job/kzg.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    let x_0_value = fetch_x_0_value_from_os_output(block_number, fetch_from_tests).await?;
+
+    // trusted setup ceremony
+    let trusted_setup_path = CURRENT_PATH.join("src/jobs/state_update_job/trusted_setup.txt");
+    let trusted_setup =
+        KzgSettings::load_trusted_setup_file(trusted_setup_path.as_path()).expect("Error loading trusted setup file");
```
apoorvsadana, line 22: this can be in `lazt_static`, we don't need to load it every time

crates/orchestrator/src/jobs/state_update_job/kzg.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    // Verifying the proof for double check
+    let eval = KzgProof::verify_kzg_proof(
+        &commitment.to_bytes(),
+        &x_0_value,
+        &y_0_value,
+        &kzg_proof.to_bytes(),
+        &trusted_setup,
+    )?;
+    assert!(eval);
```
apoorvsadana, line 36: this will panic, let's throw an error?

crates/orchestrator/src/jobs/state_update_job/kzg.rs
```diff
@@ -0,0 +1,124 @@
...
+    fetch_from_tests: Option<bool>,
+) -> color_eyre::Result<(Vec<u8>, KzgProof)> {
+    let blob_data = fetch_blob_data_for_block(block_number, fetch_from_tests).await?;
+    let mut fixed_size_blob: [u8; BYTES_PER_BLOB] = [0; BYTES_PER_BLOB];
+    fixed_size_blob.copy_from_slice(blob_data.as_slice());
+
+    let x_0_value = fetch_x_0_value_from_os_output(block_number, fetch_from_tests).await?;
+
+    // trusted setup ceremony
+    let trusted_setup_path = CURRENT_PATH.join("src/jobs/state_update_job/trusted_setup.txt");
```
apoorvsadana, line 20: there's duplicated code for trusted setup (even the file), let's move it to one place as we discussed today

crates/orchestrator/src/jobs/state_update_job/kzg.rs
```diff
@@ -0,0 +1,124 @@
...
+    fetch_from_tests: Option<bool>,
+) -> color_eyre::Result<Bytes32> {
+    let fetch_from_tests = fetch_from_tests.unwrap_or(true);
+    let x_0 = match fetch_from_tests {
+        true => {
+            let x_0_path = CURRENT_PATH.join(format!("src/jobs/state_update_job/test_data/{}/x_0.txt", block_number));
+            let data = std::fs::read_to_string(x_0_path)?;
+            Bytes32::from_hex(&data).unwrap()
+        }
+        false => unimplemented!(),
```
apoorvsadana, line 76: do we know how to fetch this now given we know the SNOS output = calldata on Eth

crates/orchestrator/src/jobs/state_update_job/mod.rs
```diff
@@ -226,10 +229,14 @@ impl StateUpdateJob {
             let onchain_data_size = 0;
             settlement_client.update_state_calldata(program_output, onchain_data_hash, onchain_data_size).await?
         } else if snos.use_kzg_da == Felt252::ONE {
-            // TODO: Build the blob & the KZG proof & send them to update_state_blobs
-            let kzg_proof = self.fetch_kzg_proof_for_block(block_no, fetch_from_tests).await;
-            let kzg_proof: [u8; 48] = kzg_proof.try_into().expect("kzg proof size must be 48 bytes");
-            settlement_client.update_state_blobs(vec![], kzg_proof).await?
+            let (blob_data, kzg_proof) = build_kzg_proof(block_no, fetch_from_tests).await?;
```
apoorvsadana, line 232: fetching from s3 can be the responsibility of this function. build_kzg_proof will just calculate the proof. it will also help in moving common code outside when we move build_proof inside the DA client

crates/orchestrator/src/jobs/state_update_job/mod.rs
```diff
@@ -253,18 +260,19 @@ impl StateUpdateJob {
...
-            true => {
-                let kzg_path =
-                    CURRENT_PATH.join(format!("src/jobs/state_update_job/test_data/{}/kzg_proof.txt", block_no));
-                std::fs::read_to_string(kzg_path).expect("Failed to read the KZG txt file").replace("0x", "")
-            }
-            false => unimplemented!("can't fetch KZG Proof from DB/S3"),
-        };
-        hex::decode(kzg_proof_str).expect("Invalid test kzg proof")
-    }
+    /// No longer needed as we are building the proof during the state update job run
```
apoorvsadana, line 263: let's delete the code then

crates/orchestrator/src/tests/jobs/da_job/mod.rs
```diff
@@ -48,7 +48,8 @@ async fn test_process_job() {
     let mut da_client = MockDaClient::new();
     let internal_id = "1";
 
-    da_client.expect_publish_state_diff().times(1).returning(|_, _| Ok(internal_id.to_string()));
+    // No longer needed step is eliminated for ethereum
```
apoorvsadana, line 51: let's delete the code

crates/orchestrator/src/tests/jobs/da_job/mod.rs
```diff
@@ -91,7 +92,7 @@ async fn test_process_job() {
             )
             .await
             .unwrap(),
-        internal_id
+        "da_job_process_done"
```
apoorvsadana, line 95: it should be the external id returned by the Da client. (txn_hash for alt_da, maybe "NA" for eth)

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -113,3 +174,102 @@ impl SettlementClient for EthereumSettlementClient {
...
+    }
+
+    Ok((sidecar_blobs, sidecar_commitments, sidecar_proofs))
+}
+
+/// Function to construct the transaction for updating the state in core contract.
+fn get_txn_input_bytes(program_output: Vec<[u8; 32]>, kzg_proof: [u8; 48]) -> Bytes {
+    let program_output_hex_string = vec_u8_32_to_hex_string(program_output);
+    let kzg_proof_hex_string = u8_48_to_hex_string(kzg_proof);
+    // cast keccak "updateStateKzgDA(uint256[] calldata programOutput, bytes calldata kzgProof)" | cut -b 1-10
```
apoorvsadana, line 208: nice job on adding these comments btw, it's good to have references on how to reproduce constants in code

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -113,3 +174,102 @@ impl SettlementClient for EthereumSettlementClient {
...
+    state_diff: &[Vec<u8>],
+    trusted_setup: &KzgSettings,
+) -> Result<(Vec<FixedBytes<131072>>, Vec<FixedBytes<48>>, Vec<FixedBytes<48>>)> {
+    let mut sidecar_blobs = vec![];
+    let mut sidecar_commitments = vec![];
+    let mut sidecar_proofs = vec![];
+
+    for blob_data in state_diff {
+        let mut fixed_size_blob: [u8; BYTES_PER_BLOB] = [0; BYTES_PER_BLOB];
+        fixed_size_blob.copy_from_slice(blob_data.as_slice());
```
apoorvsadana, line 189: same as previously mentioned

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -84,6 +99,52 @@ impl SettlementClient for EthereumSettlementClient {
         Ok(format!("0x{:x}", tx_receipt.transaction_hash))
     }
 
+    async fn update_state_blobs_and_blob(
```
apoorvsadana, line 102: `update_state_with_blobs`?

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -84,6 +99,52 @@ impl SettlementClient for EthereumSettlementClient {
...
+        program_output: Vec<[u8; 32]>,
+        kzg_proof: [u8; 48],
+        state_diff: Vec<Vec<u8>>,
+    ) -> Result<String> {
+        let trusted_setup = KzgSettings::load_trusted_setup_file(Path::new("./trusted_setup.txt"))
+            .expect("issue while loading the trusted setup");
+        let (sidecar_blobs, sidecar_commitments, sidecar_proofs) = prepare_sidecar(&state_diff, &trusted_setup).await?;
+        let sidecar = BlobTransactionSidecar::new(sidecar_blobs, sidecar_commitments, sidecar_proofs);
+
+        let eip1559_est = self.provider.estimate_eip1559_fees(None).await?;
```
apoorvsadana, line 113: but this is a 4844 txn, won't this be incorrect?

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -113,3 +174,102 @@ impl SettlementClient for EthereumSettlementClient {
...
+fn get_txn_input_bytes(program_output: Vec<[u8; 32]>, kzg_proof: [u8; 48]) -> Bytes {
+    let program_output_hex_string = vec_u8_32_to_hex_string(program_output);
+    let kzg_proof_hex_string = u8_48_to_hex_string(kzg_proof);
+    // cast keccak "updateStateKzgDA(uint256[] calldata programOutput, bytes calldata kzgProof)" | cut -b 1-10
+    let function_selector = "0x1a790556";
+
+    Bytes::from(program_output_hex_string + &kzg_proof_hex_string + function_selector)
+}
+
+#[allow(clippy::format_collect)]
```
apoorvsadana, line 214: what does `format_collect` do?

crates/orchestrator/src/jobs/state_update_job/kzg.rs
ocdbytes: so in our case when SNOS run is complete then we will store the output and we store x_0 point into another file. wdyt ?

crates/settlement-clients/ethereum/src/lib.rs
ocdbytes: no, so eip 1559 is gas fee calculating standard of etheruem.

crates/settlement-clients/ethereum/src/lib.rs
ocdbytes: clippy me error aa rha tha to ignore krne ke lie dala h

crates/orchestrator/src/tests/jobs/da_job/mod.rs
ocdbytes: returning `0xbeef` for eth

crates/da-clients/ethereum/src/lib.rs
```diff
@@ -30,57 +19,60 @@ pub mod config;
...
-        let signature = wallet.sign_transaction(&mut variant).await?;
-        let tx_signed = variant.into_signed(signature);
-        let tx_envelope: TxEnvelope = tx_signed.into();
-        let encoded = tx_envelope.encoded_2718();
-
-        let pending_tx = provider.send_raw_transaction(&encoded).await?;
-
-        Ok(pending_tx.tx_hash().to_string())
+    async fn publish_state_diff(&self, _state_diff: Vec<Vec<u8>>, _to: &[u8; 32]) -> Result<String> {
+        // Code no longer needed in Ethereum DA client because
```
apoorvsadana, line 28: We can remove it then. We almost never need to keep commented code in a repo. (check https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

crates/da-clients/ethereum/src/lib.rs
```diff
@@ -30,57 +19,60 @@ pub mod config;
...
+        // let signature = wallet.sign_transaction(&mut variant).await?;
+        // let tx_signed = variant.into_signed(signature);
+        // let tx_envelope: TxEnvelope = tx_signed.into();
+        // let encoded = tx_envelope.encoded_2718();
+        //
+        // let pending_tx = provider.send_raw_transaction(&encoded).await?;
+        //
+        // Ok(pending_tx.tx_hash().to_string())
+
+        Ok("0xbeef".to_string())
```
apoorvsadana, line 75: should we return NA instead?

crates/da-clients/ethereum/src/lib.rs
```diff
@@ -30,57 +19,60 @@ pub mod config;
...
     }
 
     async fn verify_inclusion(&self, external_id: &str) -> Result<DaVerificationStatus> {
```
apoorvsadana, line 78: this would change as well right? we don't need to verify inclusion for anything now?

crates/orchestrator/src/config.rs
```diff
@@ -161,3 +175,10 @@ async fn build_settlement_client(settings_provider: &impl SettingsProvider) -> B
         _ => panic!("Unsupported Settlement layer"),
     }
 }
+
+async fn build_storage_client() -> Box<dyn DataStorage + Send + Sync> {
+    match get_env_var_or_panic("DATA_STORAGE").as_str() {
+        "aws" => Box::new(AWSS3::new(AWSS3Config::new_from_env()).await),
```
apoorvsadana, line 181: should we make this s3? aws is not self explanatory

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -272,9 +273,29 @@ async fn state_update_to_blob_data(
         blob_data.push(*compiled_class_hash);
     }
 
+    // saving the blob data of the block to S3 bucket
+    store_blob_data_s3(blob_data.clone(), block_no, config).await?;
```
apoorvsadana, line 277: we shouldn't use `s3` anywhere outside the s3 crate, the user could be using some other storage client. we can just call it `store_blob_data`

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -272,9 +273,29 @@ async fn state_update_to_blob_data(
...
+
     Ok(blob_data)
 }
 
+/// To store the blob data in S3 bucket with path <block_number>/blob_data.txt
+async fn store_blob_data_s3(blob_data: Vec<FieldElement>, block_number: u64, config: &Config) -> Result<()> {
```
apoorvsadana, line 283: same as above

crates/orchestrator/src/jobs/proving_job/mod.rs
```diff
@@ -41,7 +41,8 @@ impl Job for ProvingJob {
     }
 
     async fn process_job(&self, config: &Config, job: &mut JobItem) -> Result<String> {
-        // TODO: allow to donwload PIE from S3
+        println!("process job start : )");
```
apoorvsadana, line 44:
```suggestion
```

crates/orchestrator/src/jobs/state_update_job/kzg.rs
```diff
@@ -0,0 +1,129 @@
...
+lazy_static! {
+    pub static ref KZG_SETTINGS: KzgSettings = KzgSettings::load_trusted_setup_file(
+        CURRENT_PATH.join("src/jobs/state_update_job/trusted_setup.txt").as_path()
+    )
+    .expect("Error loading trusted setup file");
+}
+
+/// Build KZG proof for a given block
+pub async fn build_kzg_proof(blob_data: Vec<u8>, x_0_value: Bytes32) -> color_eyre::Result<(Vec<u8>, KzgProof)> {
+    println!("blob_data: {:?}", blob_data.len());
```
apoorvsadana, line 16:
```suggestion
```

crates/orchestrator/src/jobs/state_update_job/kzg.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    // Verifying the proof for double check
+    let eval = KzgProof::verify_kzg_proof(
+        &commitment.to_bytes(),
+        &x_0_value,
+        &y_0_value,
+        &kzg_proof.to_bytes(),
+        &KZG_SETTINGS,
+    )?;
+    assert!(eval, "ERROR : Assertion failed, not able to verify the proof.");
```
apoorvsadana, line 31: this will panic, we should return an error

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -272,9 +273,29 @@ async fn state_update_to_blob_data(
...
+    let s3_client = config.storage();
+    let key = block_number.to_string() + "/blob_data.txt";
```
apoorvsadana, line 285: `"/blob_data.txt"` should be a constant so it can be reused at the other place

apoorvsadana: same for "x_0.txt"

crates/settlement-clients/ethereum/src/lib.rs
apoorvsadana: can you check what it does

crates/da-clients/ethereum/src/lib.rs
ocdbytes: I was meaning to ask this should  I just return `DaVerificationStatus` as `Verified` for ETH DA client inclusion ?

crates/orchestrator/src/config.rs
ocdbytes: done

crates/orchestrator/src/jobs/da_job/mod.rs
ocdbytes: done

crates/orchestrator/src/jobs/state_update_job/mod.rs
```diff
@@ -203,13 +198,7 @@ impl StateUpdateJob {
...
-        &self,
-        config: &Config,
-        block_no: u64,
-        snos: StarknetOsOutput,
-        fetch_from_tests: Option<bool>,
-    ) -> Result<String> {
+    async fn update_state_for_block(&self, config: &Config, block_no: u64, snos: StarknetOsOutput) -> Result<String> {
         let starknet_client = config.starknet_client();
         let settlement_client = config.settlement_client();
         let last_tx_hash_executed = if snos.use_kzg_da == Felt252::ZERO {
```
apoorvsadana, line 204: let's make this `unimplemented!()` for now I guess

crates/orchestrator/src/jobs/state_update_job/kzg.rs
```diff
@@ -0,0 +1,135 @@
...
+        CURRENT_PATH.join("src/jobs/state_update_job/trusted_setup.txt").as_path()
+    )
+    .expect("Error loading trusted setup file");
+}
+
+pub const BLOB_DATA_FILE_NAME: &str = "blob_data.txt";
+pub const X_0_FILE_NAME: &str = "x_0.txt";
+
+/// Build KZG proof for a given block
+pub async fn build_kzg_proof(blob_data: Vec<u8>, x_0_value: Bytes32) -> color_eyre::Result<(Vec<u8>, KzgProof)> {
```
apoorvsadana, line 19: the build proof part needs to be inside the settlement client. celestia, for example, will have a completely different way of building the proof

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -113,3 +175,104 @@ impl SettlementClient for EthereumSettlementClient {
...
+async fn prepare_sidecar(
+    state_diff: &[Vec<u8>],
+    trusted_setup: &KzgSettings,
+) -> Result<(Vec<FixedBytes<131072>>, Vec<FixedBytes<48>>, Vec<FixedBytes<48>>)> {
+    let mut sidecar_blobs = vec![];
+    let mut sidecar_commitments = vec![];
+    let mut sidecar_proofs = vec![];
+
+    for blob_data in state_diff {
+        let fixed_size_blob: [u8; BYTES_PER_BLOB] = blob_data.as_slice().try_into()?;
```
apoorvsadana, line 189: we're using `BYTES_PER_BLOB` in multiple places which comes from `alloy`. this isn't the correct design. we have implemented `max_bytes_per_blob` on da clients, we should actually use that over here

crates/settlement-clients/ethereum/src/lib.rs
apoorvsadana: this is still there in da client

crates/da-clients/ethereum/src/lib.rs
```diff
@@ -30,57 +19,13 @@ pub mod config;
...
-        let signature = wallet.sign_transaction(&mut variant).await?;
-        let tx_signed = variant.into_signed(signature);
-        let tx_envelope: TxEnvelope = tx_signed.into();
-        let encoded = tx_envelope.encoded_2718();
-
-        let pending_tx = provider.send_raw_transaction(&encoded).await?;
-
-        Ok(pending_tx.tx_hash().to_string())
+    async fn publish_state_diff(&self, _state_diff: Vec<Vec<u8>>, _to: &[u8; 32]) -> Result<String> {
+        Ok("NA".to_string())
```
apoorvsadana, line 28: let's add a comment explaining why we aren't pushing the state diff here

crates/orchestrator/src/jobs/state_update_job/mod.rs
```diff
@@ -226,10 +215,18 @@ impl StateUpdateJob {
             let onchain_data_size = 0;
             settlement_client.update_state_calldata(program_output, onchain_data_hash, onchain_data_size).await?
         } else if snos.use_kzg_da == Felt252::ONE {
-            // TODO: Build the blob & the KZG proof & send them to update_state_blobs
-            let kzg_proof = self.fetch_kzg_proof_for_block(block_no, fetch_from_tests).await;
-            let kzg_proof: [u8; 48] = kzg_proof.try_into().expect("kzg proof size must be 48 bytes");
-            settlement_client.update_state_blobs(vec![], kzg_proof).await?
+            let blob_data = fetch_blob_data_for_block(block_no).await?;
+            let x_0_value = fetch_x_0_value_from_os_output(block_no).await?;
```
apoorvsadana, line 219: Now I actually think this will become `fetch_snos_output`. It doesn't make sense to only store x0 in a file. wdyt?

apoorvsadana:
Can you also move `test_data` from `update_state` job to the tests folder. Miss from my side, it came in some older PR

crates/orchestrator/src/tests/jobs/state_update_job/mod.rs
```diff
@@ -51,30 +58,58 @@ async fn test_process_job() {
         let program_output: Vec<[u8; 32]> = vec![];
```
apoorvsadana, line 58: I know it's not in this PR but instead of looping over the blocks, let's make these different test cases?

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -76,6 +79,7 @@ impl Job for DaJob {
             MaybePendingStateUpdate::Update(state_update) => state_update,
         };
         // constructing the data from the rpc
+        // TODO: Need to update the test_mode variable to false after implementation
```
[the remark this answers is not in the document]
ocdbytes: removed old comment.

crates/settlement-clients/ethereum/src/lib.rs
ocdbytes: same as above we need to refactor a lot of code for this to happen. Will discuss this tomorrow in office.

crates/orchestrator/src/jobs/state_update_job/mod.rs
```diff
@@ -226,10 +215,18 @@ impl StateUpdateJob {
...
+
+            let (blob_data, kzg_proof) = build_kzg_proof(blob_data, x_0_value).await?;
+
+            // Temp solution for formatting to blob
+            // TODO : Need to implement solution to handle multiple vectors (multiple data) as It will be updated in the upcoming starknet update.
```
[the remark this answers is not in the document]
ocdbytes: store kr skte h but dekhna padega ki how we will handle it in our code if there are multiple blobs in the vector. So code changes hoenge kaafi jagah so after we change the client code then only we can move to this issue.

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -272,9 +275,29 @@ async fn state_update_to_blob_data(
...
 
+/// To store the blob data using the storage client with path <block_number>/blob_data.txt
+async fn store_blob_data(blob_data: Vec<FieldElement>, block_number: u64, config: &Config) -> Result<()> {
+    let s3_client = config.storage();
+    let key = block_number.to_string() + "/" + BLOB_DATA_FILE_NAME;
+    let data_blob_big_uint = convert_to_biguint(blob_data.clone());
+
+    // TODO : Figure out the approach when there are multiple blobs in blobs_array
+    let blobs_array = data_to_blobs(config.da_client().max_bytes_per_blob().await, data_blob_big_uint).unwrap();
+    let blob = blobs_array[0].clone();
```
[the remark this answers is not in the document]
ocdbytes: ok we will implement that.

crates/orchestrator/src/jobs/state_update_job/mod.rs
ocdbytes: I think it is easier to read if just store the x_0 otherwise we would need to go line by line and at certain index read the x_0 point. So I was thinking that what if we store the x_0 separately after SNOS run.

crates/orchestrator/src/tests/jobs/state_update_job/mod.rs
ocdbytes: Ig ese hi theek h vese bhi mock test h idts abhi ke lie necessary hoega. Vese bhi we are going to add more robust test cases ig uske baad update ho hi jaega as per the requirements.

crates/orchestrator/src/jobs/state_update_job/kzg.rs
apoorvsadana: that's fine, the config is available everywhere in the app so it can access the storage client right?

crates/settlement-clients/ethereum/src/lib.rs
apoorvsadana: hmm, why can't we just get the da client from config and call max bytes per blob? can discuss in office also :)

crates/orchestrator/src/jobs/state_update_job/mod.rs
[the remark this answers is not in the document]
apoorvsadana: ok. what would be the code changes exactly? also, maybe we can use `bincode` crate for serializing and deserializing, not sure

crates/orchestrator/src/jobs/state_update_job/mod.rs
apoorvsadana: hmm. so i was thinking we do the parsing and everything at the SNOS job itself (the same way we assume it right now) and store a json which has fields like 

```json
{
  "pie": <PIE>,
  "x0": x0
}
```

apoorvsadana: basically, avoiding multiple files

crates/orchestrator/src/tests/jobs/state_update_job/mod.rs
apoorvsadana: hmm, i think we should move it to different cases but we can do it another PR for sure. regarding the more robust cases, ya we definitely need to decide an exhaustive list of test cases needed but these tests would still remain

crates/orchestrator/src/jobs/state_update_job/mod.rs
ocdbytes: this approach also works.

PR #46, opened 2024-07-20
Build Client inside DaConfig

heemankv:
This PR implements `build_client` inside `DaConfig` and removes need for implementing `tryfrom` / `from` completely.

### Problem it solves :
Not all DA clients have a `sync` da client support, this PR allows creating the DA clients `async`.

### Important points to evaluate: 
- declaration of `build_client` on DaConfig.
- `ethereum` DA setup in `build_da_client` inside `crates/orchestrator/src/config.rs`.
- `get_env_var_optional ` : optional env support, if `env` variable is not found, `None` is passed.

### DIFF
--- a/crates/utils/src/env_utils.rs
+++ b/crates/utils/src/env_utils.rs
@@ -1,27 +1,19 @@
 use std::env::VarError;
 
-
 pub fn get_env_var(key: &str) -> Result<String, VarError> {
     std::env::var(key)
 }
 
-pub fn get_env_var_optional(key: &str) -> Result<Option<String>,VarError> {
-    match get_env_var(key){
-        Ok(s) => {
-            Ok(Some(s))
-        }
-        Err(VarError::NotPresent) => {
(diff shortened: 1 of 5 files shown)


crates/utils/src/env_utils.rs
```diff
@@ -1,7 +1,19 @@
...
+
+pub fn get_env_var_optional(key: &str) -> Result<Option<String>, VarError> {
+    match get_env_var(key) {
+        Ok(s) => Ok(Some(s)),
+        Err(VarError::NotPresent) => Ok(None),
+        Err(e) => Err(e),
+    }
+}
+
+pub fn get_env_car_optional_or_panic(key: &str) -> Option<String> {
```
apoorvsadana, line 15:
```suggestion
pub fn get_env_var_optional_or_panic(key: &str) -> Option<String> {
```

apoorvsadana (approved):

Issue #11, opened 2024-06-01
implement SNOS job


apoorvsadana:
Complete the SNOS job implementation over here - https://github.com/karnotxyz/madara-orchestrator/blob/2a9d77b7188afadefe3feba3347a909fc83299ed/crates/orchestrator/src/jobs/snos_job/mod.rs#L12

1. We plan to add an endpoint on Madara called `getSnosInput`. For now, we can assume this endpoint exists and we can call it to get the SNOS input - https://github.com/keep-starknet-strange/snos/blob/da26557078111b5d489154b162f3f810174f5b69/src/io/input.rs#L19
2. Use this input to execute the SNOS and get the PIE
3. Store the PIE in a database/S3. I think storing in S3 might make more sense as we won't really need to query the database
4. To add S3 support, we can add a generalised trait like we do for `database` and `queues` already

sBaki92:
Hi, I would like to take it :)

apoorvsadana:
Assigned!

ocdbytes:
Assign this to me : )

akhercha:
gm, I can take this one if it's free!

apoorvsadana:
Yes, assigned!

akhercha:
Done in https://github.com/madara-alliance/madara-orchestrator/pull/136

PR #50, opened 2024-07-22
[paused] feat: snos job

akhercha:
## ⏸ On hold

This PR is currently paused until the [snos](https://github.com/keep-starknet-strange/snos) repository has been updated (since the way to run Snos in the context of Madara will change a lot and we don't have any spec/design yet).

________

Resolves: #11 

## Added

- Implementation of the `SnosJob`
- `StateJob::DummyState` - used as a replacement until we can import the structure `BlockifierStateAdapter` from Madara. Currently not possible dues to conflicting dependencies versions between Snos <=> Madara <=> Cairo-vm. (It may not be needed at all to import that depending on the next updates from snos, their goal is to make it easier to run the Snos in the context of Madara)
- `HttpRpcClient` to our `Config` for RPC requests to L1/Madara (gas price & snos input at the moment)
- `conversions` & `time` utils functions


crates/orchestrator/src/jobs/snos_job/mod.rs
```diff
@@ -29,28 +51,133 @@ impl Job for SnosJob {
...
 
     fn max_process_attempts(&self) -> u64 {
-        todo!()
+        1
     }
 
     fn max_verification_attempts(&self) -> u64 {
-        todo!()
+        // TODO: isn't 10 a lot?
+        10
```
apoorvsadana, line 148: ya if there's no verification then 10 would be a lot :)

crates/orchestrator/src/jobs/snos_job/mod.rs
```diff
@@ -29,28 +51,133 @@ impl Job for SnosJob {
...
     }
 
     fn verification_polling_delay_seconds(&self) -> u64 {
-        todo!()
+        // TODO: what is an average run time for SNOS?
+        60
```
apoorvsadana, line 153: we don't know the runtime yet but in the current implementation looks like we run in synchronously so verification delay can be like 1s

akhercha: Ah yes true - updated

Tranduy1dol:
Hi @apoorvsadana , I just want to know if there are any plans for SNOS. I've noticed that this PR has been on hold for a while.

akhercha:
> Hi @apoorvsadana , I just want to know if there are any plans for SNOS. I've noticed that this PR has been on hold for a while.

Hey @Tranduy1dol - this PR is on hold until the `snos` repository is ready to be integrated with Madara. Until then, we can't really continue this work.

Issue #37, opened 2024-07-18
feat: add DA worker


apoorvsadana:
We previously planned to avoid a DA worker for L2s using Ethereum because the blob is submitted directly during state update. However, we now plan to add this back because it allows for a more generalised design across Alt DA layers and Ethereum. The DA worker will

1. Fetch the last job for which we completed DA
2. Create a DA job for all blocks > last job and <= latest completed block on the chain

PR #51, opened 2024-07-22
feat: Data Submission Worker Integration.

heemankv:
This PR resolves Issue #37.

### Logic : 
- We assume absolute non sequential calls, no need to wait on previous data submission jobs to be completed before triggering for new block.
- `Y` : Take last completed Proof Job.
- `X` : Take last triggered Data Submission Job.
- `X <= Y` 
- Trigger data submission jobs for all `Y - X` jobs.


crates/orchestrator/src/workers/data_submission.rs
```diff
@@ -0,0 +1,55 @@
...
+    // 1. Fetch the latest completed Proving job.
+    // 2. Fetch the latest DA job creation.
+    // 3. Create jobs from after the lastest DA job already created till latest completed proving job.
+    async fn run_worker(&self) -> Result<(), Box<dyn Error>> {
+        let config = config().await;
+
+        // provides latest completed proof creation job id
+        let latest_proven_job_id = config
+            .database()
+            .get_last_successful_job_by_type(JobType::ProofCreation)
```
ocdbytes, line 23: We can delete this function `get_last_successful_job_by_type` from the mongo db functions. Instead we can use `get_last_successful_job_by_type` that is doing the same thing.

crates/orchestrator/src/workers/data_submission.rs
```diff
@@ -0,0 +1,55 @@
...
+            .await
+            .unwrap()
+            .map(|item| item.internal_id)
+            .unwrap_or("0".to_string());
+
+        // provides latest triggered data submission job id
+        let latest_data_submission_job_id = config
+            .database()
+            .get_latest_job_by_type_and_internal_id(JobType::DataSubmission)
```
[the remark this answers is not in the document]
ocdbytes: same as above we can remove this function and use `get_last_successful_job_by_type`

crates/orchestrator/src/workers/data_submission.rs
```diff
@@ -0,0 +1,55 @@
...
+
+        let latest_data_submission_id: u64 = latest_data_submission_job_id.parse()?;
+        let latest_proven_id: u64 = latest_proven_job_id.parse()?;
+
+        let block_diff = latest_proven_id - latest_data_submission_id;
+
+        // if all blocks are processed
+        if block_diff == 0 {
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
heemankv: Valid, will remove, thanks

crates/orchestrator/src/workers/data_submission.rs
```diff
@@ -0,0 +1,55 @@
...
+pub struct DataSubmissionWorker;
+
+#[async_trait]
+impl Worker for DataSubmissionWorker {
+    // 0. All ids are assumed to be block numbers.
+    // 1. Fetch the latest completed Proving job.
+    // 2. Fetch the latest DA job creation.
+    // 3. Create jobs from after the lastest DA job already created till latest completed proving job.
+    async fn run_worker(&self) -> Result<(), Box<dyn Error>> {
+        let config = config().await;
```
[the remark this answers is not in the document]
heemankv: Valid, being implemented.

crates/orchestrator/src/workers/data_submission.rs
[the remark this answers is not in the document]
heemankv: - changed the name `get_latest_job_by_type_and_internal_id` to `get_latest_job_by_type`.
- kept function definition as it is. 
- for `latest_data_submission_job_id` we want the last triggered job hence we use `get_latest_job_by_type`, whereas for `latest_proven_job_id` we need last Completed job and hence we use `get_last_successful_job_by_type` both of which serves different use cases.

crates/orchestrator/src/workers/snos.rs
heemankv, line 1: There is a discrepancy in what inline `Git Lens` extension and Github shows here.

Screenshot 2024-07-23 at 2 11 15 PM

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## Added
 
+- creation of data submission jobs.
```
apoorvsadana, line 9:
```suggestion
- implemented DA worker
```

crates/orchestrator/src/jobs/mod.rs
```diff
@@ -134,8 +134,9 @@ pub async fn verify_job(id: Uuid) -> Result<()> {
         JobVerificationStatus::Verified => {
             config.database().update_job_status(&job, JobStatus::Completed).await?;
         }
-        JobVerificationStatus::Rejected(e) => {
-            config.database().update_job_status(&job, JobStatus::VerificationFailed(e)).await?;
+        JobVerificationStatus::Rejected(_) => {
+            // TODO: change '_' to 'e' and add error 'e' to metadata of job status.
```
apoorvsadana, line 138: do we need a TODO, can we do it here? Also, let's do a `log::error!` to log `e`

crates/orchestrator/src/workers/mod.rs
```diff
@@ -1,13 +1,39 @@
...
 
+pub mod data_submission;
 pub mod proof_registration;
 pub mod proving;
 pub mod snos;
 pub mod update_state;
 
 #[async_trait]
 pub trait Worker: Send + Sync {
+    async fn run_worker_if_enabled(&self) -> Result<(), Box<dyn Error>> {
```
apoorvsadana, line 13: interesting, do you see a use case to have a separate function for this?

crates/orchestrator/src/workers/mod.rs
```diff
@@ -1,13 +1,39 @@
...
+    // TODO: Assumption : False Negative
+    // we are assuming that the worker will spawn only 1 job for a block and no two jobs will ever exist
+    // for a single block, the code might fail to work as expected if this happens.
+
+    // Checks if any of the jobs have failed
+    // Haults any new job creation till all the count of failed jobs is not Zero.
+    async fn is_worker_enabled(&self) -> Result<bool, Box<dyn Error>> {
+        let config = config().await;
+
+        let failed_da_jobs = config.database().get_jobs_by_status(JobStatus::VerificationFailed).await?;
```
apoorvsadana, line 31: getting all jobs by status is very large query, this will become very slow with time. let's do a limit 1 query

crates/orchestrator/src/workers/data_submission.rs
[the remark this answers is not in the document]
apoorvsadana: I think we can have just 2 functions - `get_latest_job_by_type` and `get_latest_job_by_type_and_status`. `get_last_successful_job_by_type` is a special case of the second

crates/orchestrator/src/workers/snos.rs
apoorvsadana: GitLens is showing the diff between the current code and the previous commit I think. Github is showing the diff with the main branch

apoorvsadana (commented):
We haven't added test cases for this code

crates/orchestrator/src/workers/mod.rs
heemankv: Allows for generalised guard clause checks, no need to rewrite same code on every worker.

crates/orchestrator/src/workers/snos.rs
heemankv: makes sense.

crates/orchestrator/src/database/mongodb/mod.rs
```diff
@@ -290,4 +294,34 @@ impl Database for MongoDb {
...
+        while let Some(result) = jobs.next().await {
+            match result {
+                Ok(job_item) => {
+                    results.push(job_item);
+                }
+                Err(e) => return Err(e.into()),
+            }
+        }
+
+        Ok(results)
```
apoorvsadana, line 325:
```rust
        let jobs: Vec<JobItem> = self
            .get_job_collection()
            .find(filter, find_options)
            .await
            .expect("Failed to fetch jobs by given job type and status")
            .try_collect()
            .await?;

        Ok(jobs)
```

this looks like it would do the job (took help from Claude :)). if it works, we can use it above as well

crates/orchestrator/src/jobs/mod.rs
```diff
@@ -135,7 +135,12 @@ pub async fn verify_job(id: Uuid) -> Result<()> {
             config.database().update_job_status(&job, JobStatus::Completed).await?;
         }
         JobVerificationStatus::Rejected(e) => {
-            config.database().update_job_status(&job, JobStatus::VerificationFailed(e)).await?;
+            config.database().update_job_status(&job, JobStatus::VerificationFailed).await?;
+            let mut metadata = job.metadata.clone();
+            metadata.insert("error".to_string(), e);
+            config.database().update_metadata(&job, metadata).await?;
```
apoorvsadana, line 141: we can have one `update_job`

crates/orchestrator/src/workers/mod.rs
```diff
@@ -1,13 +1,39 @@
...
+    async fn run_worker_if_enabled(&self) -> Result<(), Box<dyn Error>> {
+        if !self.is_worker_enabled().await? {
+            return Ok(());
+        }
+        self.run_worker().await
+    }
+
     async fn run_worker(&self) -> Result<(), Box<dyn Error>>;
+
+    // TODO: Assumption : False Negative
```
apoorvsadana, line 22: is this a TODO or just an assumption?

crates/orchestrator/src/workers/snos.rs
```diff
@@ -22,7 +22,7 @@ impl Worker for SnosWorker {
         let latest_block_number = provider.block_number().await?;
         let latest_block_processed_data = config
             .database()
-            .get_last_successful_job_by_type(JobType::SnosRun)
+            .get_latest_job_by_type_and_status(JobType::ProofCreation, JobStatus::Completed)
```
apoorvsadana, line 25: is this supposed to be SnosRun?

crates/orchestrator/src/database/mongodb/mod.rs
```diff
@@ -245,18 +246,22 @@ impl Database for MongoDb {
...
-            "job_status": bson::to_bson(&JobStatus::Completed)?
+            "job_status": bson::to_bson(&job_status)?
         };
         let find_options = FindOneOptions::builder().sort(doc! { "internal_id": -1 }).build();
 
         Ok(self
             .get_job_collection()
             .find_one(filter, find_options)
             .await
-            .expect("Failed to fetch latest job by given job type"))
```
apoorvsadana, line 259: expect would panic, miss from our side, let's remove all expects in the mongodb code

crates/orchestrator/src/database/mongodb/mod.rs
```diff
@@ -290,4 +295,19 @@ impl Database for MongoDb {
...
+        let filter = doc! {
+            "job_status": bson::to_bson(&job_status)?
+        };
+
+        let mut find_options = None;
+        if let Some(val) = limit {
+            find_options = Some(FindOptions::builder().limit(Some(val)).build())
+        };
+
+        let jobs = self.get_job_collection().find(filter, find_options).await?.try_collect().await?;
```
apoorvsadana, line 309: nice, looks like this works, can we do this above as well?

crates/orchestrator/src/jobs/mod.rs
```diff
@@ -135,7 +135,13 @@ pub async fn verify_job(id: Uuid) -> Result<()> {
...
         }
         JobVerificationStatus::Rejected(e) => {
-            config.database().update_job_status(&job, JobStatus::VerificationFailed(e)).await?;
+            let mut new_job = job.clone();
+            new_job.metadata.insert("error".to_string(), e);
+            new_job.status = JobStatus::VerificationFailed;
+
+            config.database().update_job(&new_job).await?;
+
+            log::error!("Invalid status {:?} for job with id {:?}. Cannot verify.", id, job.status);
```
apoorvsadana, line 144:
```suggestion
            log::error!("Verification failed for job with id {:?}. Cannot verify.", id);
```

crates/orchestrator/src/workers/data_submission.rs
```diff
@@ -0,0 +1,48 @@
...
+            .await
+            .unwrap()
+            .map(|item| item.internal_id)
+            .unwrap_or("0".to_string());
+
+        let latest_data_submission_id: u64 = latest_data_submission_job_id.parse()?;
+        let latest_proven_id: u64 = latest_proven_job_id.parse()?;
+
+        // creating data submission jobs for latest blocks without pre-running data submission jobs jobs don't yet exist.
+        for x in latest_data_submission_id + 1..latest_proven_id + 1 {
```
akhercha, line 42: `x` should be a meaningful name, ie `new_job_id`

crates/orchestrator/src/workers/data_submission.rs
```diff
@@ -0,0 +1,48 @@
...
+            .get_latest_job_by_type_and_status(JobType::ProofCreation, JobStatus::Completed)
+            .await
+            .unwrap()
+            .map(|item| item.internal_id)
+            .unwrap_or("0".to_string());
+
+        // provides latest triggered data submission job id
+        let latest_data_submission_job_id = config
+            .database()
+            .get_latest_job_by_type(JobType::DataSubmission)
```
akhercha, line 32: We don't check by `status` here too? How do we handle failed jobs?

akhercha: Nevermind, just saw the `is_worker_enabled`, it should not happen.

crates/orchestrator/src/workers/data_submission.rs
akhercha, line 42: `x` should be a meaningful name like `new_job_id`

crates/orchestrator/src/workers/mod.rs
```diff
@@ -1,13 +1,39 @@
...
+    // Assumption : False Negative
+    // we are assuming that the worker will spawn only 1 job for a block and no two jobs will ever exist
+    // for a single block, the code might fail to work as expected if this happens.
+
+    // Checks if any of the jobs have failed
+    // Haults any new job creation till all the count of failed jobs is not Zero.
+    async fn is_worker_enabled(&self) -> Result<bool, Box<dyn Error>> {
+        let config = config().await;
+
+        let failed_da_jobs = config.database().get_jobs_by_status(JobStatus::VerificationFailed, Some(1)).await?;
```
akhercha, line 31: What about pending jobs?
If we have
`[DONE, DONE, DONE, PENDING]` as an array of 4 jobs,

and we start the worker:
`[DONE, DONE, DONE, PENDING, CREATED]`

and then the pending job fails. It won't be handled at the moment right?

crates/orchestrator/src/workers/data_submission.rs
```diff
@@ -0,0 +1,48 @@
...
+            .await
+            .unwrap()
+            .map(|item| item.internal_id)
+            .unwrap_or("0".to_string());
```
EvolveArt, line 36: shouldn't we do nothing if we dont have any DA job ?

crates/orchestrator/src/workers/data_submission.rs
```diff
@@ -0,0 +1,48 @@
...
+
+        let latest_data_submission_id: u64 = latest_data_submission_job_id.parse()?;
+        let latest_proven_id: u64 = latest_proven_job_id.parse()?;
+
+        // creating data submission jobs for latest blocks without pre-running data submission jobs jobs don't yet exist.
```
EvolveArt, line 41: comment not clear

crates/orchestrator/src/workers/mod.rs
```diff
@@ -1,13 +1,39 @@
...
+        if !self.is_worker_enabled().await? {
+            return Ok(());
+        }
+        self.run_worker().await
+    }
+
     async fn run_worker(&self) -> Result<(), Box<dyn Error>>;
+
+    // Assumption : False Negative
+    // we are assuming that the worker will spawn only 1 job for a block and no two jobs will ever exist
```
EvolveArt, line 23: im not 100% sure of this assumption, like jobs in my mind should be immutable. It is quite error prone to do the opposite.

crates/orchestrator/src/workers/data_submission.rs
```diff
@@ -0,0 +1,48 @@
...
+use crate::jobs::types::{JobStatus, JobType};
+use crate::workers::Worker;
+use async_trait::async_trait;
+use std::collections::HashMap;
+use std::error::Error;
+
+pub struct DataSubmissionWorker;
+
+#[async_trait]
+impl Worker for DataSubmissionWorker {
```
EvolveArt, line 12: name is not very clear, afaiu this job's role is to make sure DA picks up proving right ? Do we never see a world where proving is ahead of DA ?

crates/orchestrator/src/workers/mod.rs
akhercha: Are we fine with that? If yes and it is handled somewhere else, can we add a comment explaining it?

crates/orchestrator/src/workers/mod.rs
EvolveArt: for me once a job has failed, it's a failed state and the only way ahead is to spawn a new one, wdyt @apoorvsadana

crates/orchestrator/src/workers/data_submission.rs
EvolveArt: yes but let's say DA jobs have failed then in that case proving is ahead ?
I think `DataAvailabilitySynchronizer` is a better name?

crates/orchestrator/src/workers/data_submission.rs
EvolveArt: for existing data submission jobs maybe instead of `pre-running` ?

crates/orchestrator/src/workers/mod.rs
heemankv: For now, yes we are fine with that.
`We assume the system to keep working till a job hasn't failed, as soon as it fails we currently halt any more execution and wait for manual intervention.`
Adding this comment.

crates/orchestrator/src/workers/data_submission.rs
heemankv: Valid, updated.

crates/orchestrator/src/workers/data_submission.rs
heemankv: @apoorvsadana ,
Note: name changed to `DataAvailabilitySynchronizer`.

crates/orchestrator/src/workers/mod.rs
apoorvsadana: So creating a new job as a whole might be difficult because we don't know at what stage the job failed. Currently, I was thinking we go ahead with this -> investigate what sort of failures we're getting and add new code accordingly to retry failed jobs (so you don't need to manually edit the DB but we add code on demand to handle failed cases based on what we see)

crates/orchestrator/src/workers/data_submission.rs
apoorvsadana: > yes but let's say DA jobs have failed then in that case proving is ahead ?

So I guess there's some confusion. Proving is always ahead. Only after a proving job is succesful we actually create a new job for DA. I would still like to keep the Worker name actually because its coherent with all other workers. wdyt

crates/orchestrator/src/workers/mod.rs
EvolveArt: yes I think that's already a better alternative, would be great to have clear error codes listed

crates/orchestrator/src/workers/data_submission.rs
EvolveArt: yes confusion cleared, agreed lets keep it that way for now although it does not submit only for one block but its role is to keep DA and proving synchronized

apoorvsadana (approved):

PR #52, opened 2024-07-23
Feat : add testing fixtures and a basic test

ocdbytes:
- Added local stack s3 bucket mocking tests for putting and getting the data.


crates/orchestrator/src/tests/controllers/mod.rs
apoorvsadana, line 1: we can delete this test case as well

crates/orchestrator/src/tests/database/mod.rs
```diff
@@ -1,7 +1,73 @@
...
+    assert_eq!(get_job_2, job_vec[1].clone());
+    assert_eq!(get_job_3, job_vec[2].clone());
+
+    Ok(())
+}
+
+// Test Util Functions
+// ==========================================
+
+fn get_random_job_item(job_type: JobType, job_status: JobStatus, internal_id: u64) -> JobItem {
```
apoorvsadana, line 63:
```suggestion
fn build_job_item(job_type: JobType, job_status: JobStatus, internal_id: u64) -> JobItem {
```

crates/orchestrator/src/tests/database/mod.rs
```diff
@@ -1,7 +1,73 @@
...
+    }
+
+    Ok(())
+}
+
+/// Tests for `create_job` operation in database trait.
+/// Creates 3 jobs and asserts them.
+#[rstest]
+#[tokio::test]
+async fn test_database_create_job() -> color_eyre::Result<()> {
```
apoorvsadana, line 24: I don't think this is the correct use of fixture, check this - https://docs.rs/rstest/latest/rstest/attr.fixture.html#async

.env.test.example
apoorvsadana, line 1: it's not an example, it's the actual test env file, it should have the values as well so anybody can do the testing

.github/workflows/coverage.yml
```diff
@@ -26,6 +41,22 @@ jobs:
...
+          echo 'ETHEREUM_RPC_URL="http://localhost:3001"' >> .env.test
+          echo 'MEMORY_PAGES_CONTRACT_ADDRESS="0x000000000000000000000000000000000001dead"' >> .env.test
+          echo 'PRIVATE_KEY="0xdead"' >> .env.test
+          echo 'ETHEREUM_PRIVATE_KEY="0x000000000000000000000000000000000000000000000000000000000000beef"' >> .env.test
+          echo 'STARKNET_SOLIDITY_CORE_CONTRACT_ADDRESS="0x000000000000000000000000000000000002dead"' >> .env.test
+          echo 'DA_LAYER="ethereum"' >> .env.test
+          echo 'PROVER_SERVICE="sharp"' >> .env.test
+          echo 'SETTLEMENT_LAYER="ethereum"' >> .env.test
+          echo 'DATA_STORAGE="s3"' >> .env.test
+          echo 'MONGODB_CONNECTION_STRING="mongodb://localhost:27017"' >> .env.test
```
apoorvsadana, line 59: move these to file itself so we can avoid making changes to both the CI and the .env

.gitignore
```diff
@@ -2,6 +2,7 @@
 .env
 .idea
 .DS_Store
+.env.test
```
apoorvsadana, line 5: we can remove this then

crates/orchestrator/src/tests/database/mod.rs
apoorvsadana: this is still not fixed

apoorvsadana (approved):

crates/orchestrator/src/config.rs
```diff
@@ -159,25 +159,140 @@ async fn build_da_client() -> Box<dyn DaClient + Send + Sync> {
...
 
-async fn build_storage_client() -> Box<dyn DataStorage + Send + Sync> {
+pub async fn build_storage_client() -> Box<dyn DataStorage + Send + Sync> {
     match get_env_var_or_panic("DATA_STORAGE").as_str() {
         "s3" => Box::new(AWSS3::new(AWSS3Config::new_from_env()).await),
         _ => panic!("Unsupported Storage Client"),
     }
 }
+
+#[cfg(test)]
```
odesenfans, line 187: Nitpick: why don't you put this in a dedicated module to avoid annotating every item with `#[cfg(test)]` ?

crates/orchestrator/src/config.rs
```diff
@@ -159,25 +159,140 @@ async fn build_da_client() -> Box<dyn DaClient + Send + Sync> {
...
+
+        // return config and server as tuple
+        let config = Config::new(
+            self.starknet_client.unwrap(),
+            self.da_client.unwrap(),
+            self.prover_client.unwrap(),
+            self.settlement_client.unwrap(),
+            self.database.unwrap(),
+            self.queue.unwrap(),
+            self.storage.unwrap(),
```
odesenfans, line 291: To avoid these unwraps, a better pattern is to do, for each entry (example with `prover_client`):

```
let prover_client = self.prover_client.unwrap_or_else(|| build_prover_service(&settings_provider));
```

and then just pass `prover_client` to `Config::new()`.

crates/orchestrator/src/data_storage/aws_s3/config.rs
```diff
@@ -12,6 +12,9 @@ pub struct AWSS3Config {
     pub s3_bucket_name: String,
     /// S3 Bucket region
     pub s3_bucket_region: String,
+    /// Endpoint url
+    #[cfg(test)]
+    pub endpoint_url: String,
```
odesenfans, line 17: Why is this only used in tests? Is there no equivalent notion when connecting to AWS directly? It would be better to avoid this difference between production and testing, if possible.

crates/orchestrator/src/database/mongodb/mod.rs
```diff
@@ -40,6 +40,10 @@ impl MongoDb {
         MongoDb { client }
     }
 
+    pub fn client(&self) -> Client {
+        self.client.clone()
```
odesenfans, line 44: Nitpick: when I see the call to `clone()`, my first reflex is to check the cost of this operation. I looked at the docs of MongoDB and saw that `Client` uses an `Arc` internally, so this is fine. But it is worth adding a comment explaining why this is fine, because for non Arc-based types doing

```
pub fn client(&self) -> &Client {
  &self.client
}
```

and then letting the caller decide if he wants to clone is better.

TL;DR, add a comment to explain that `Client` uses `Arc` and therefore is cheap to clone.

crates/orchestrator/src/routes.rs
```diff
@@ -1,16 +1,10 @@
...
 
-use crate::controllers::jobs_controller;
-
 pub fn app_router() -> Router {
-    Router::new()
-        .route("/health", get(root))
-        .nest("/v1/dev", dev_routes())
-        .nest("/v1/job", job_routes())
-        .fallback(handler_404)
+    Router::new().route("/health", get(root)).nest("/v1/dev", dev_routes()).fallback(handler_404)
```
odesenfans, line 7: How is this change related to the PR?

crates/orchestrator/src/tests/common/mod.rs
```diff
@@ -74,3 +77,10 @@ pub fn custom_job_item(default_job_item: JobItem, #[default(String::from("0"))]
 
     job_item
 }
+
+pub async fn drop_database() -> color_eyre::Result<()> {
+    let db_client: Client = MongoDb::new(MongoDbConfig::new_from_env()).await.client();
+    // dropping `jobs` collection.
+    db_client.database("orchestrator").collection::<JobItem>("jobs").drop(None).await?;
```
odesenfans, line 84: Careful that you'll need to add every future collection in here as well. It might be better to just drop every collection under the `orchestrator` DB.

crates/orchestrator/src/tests/data_storage/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+use crate::data_storage::aws_s3::AWSS3;
+use crate::data_storage::{DataStorage, DataStorageConfig};
+use bytes::Bytes;
+use rstest::rstest;
+use serde_json::json;
+use utils::env_utils::get_env_var_or_panic;
+
+#[rstest]
+#[tokio::test]
+async fn test_put_and_get_data_s3() -> color_eyre::Result<()> {
```
odesenfans, line 12: So this is a sanity check to see that putting and getting to/from S3 is working, right? Don't hesitate to document what your tests intend to do, that helps for review.

crates/orchestrator/src/tests/database/mod.rs
```diff
@@ -1,7 +1,63 @@
...
+    TestConfigBuilder::new().build().await;
+
+    drop_database().await.unwrap();
```
odesenfans, line 21: Use a fixture to start with a fresh DB and an already built config.

odesenfans (changes requested):
A few comments and questions. Mostly, you should document the intent of your tests and use fixtures to clean up the DB.

crates/orchestrator/src/config.rs
heemankv: Valid, Moved this code to a `orchestrator/src/tests/config.rs` 
No need to feature flag now.

crates/orchestrator/src/database/mongodb/mod.rs
heemankv: Valid and a great catch, 
added docs to describe that `client` uses `Arc`.

crates/orchestrator/src/tests/common/mod.rs
heemankv: Valid, changed to dropping all the collections for now.
+ Added comment to how to change to specific collection.

crates/orchestrator/src/tests/data_storage/mod.rs
heemankv: Valid, Added Docs for this test.

crates/orchestrator/src/data_storage/aws_s3/config.rs
heemankv: Localstack endpoint URL : http://localhost.localstack.cloud:4566
AWS endpoint URL. : https://s3.{region}.amazonaws.com/{bucket-name}/{key}

We are using [localstack](https://hub.docker.com/r/localstack/localstack) for mocking AWS locally.
Localstack requires `endpoint_url` to get connected.

Connection to AWS doesn't necessarily require the endpoint url as shown [here](https://docs.rs/aws-config/latest/aws_config/index.html).
Rather, the connection to localstack does as shown [here](https://docs.aws.amazon.com/sdk-for-rust/latest/dg/localstack.html#:~:text=if%20use_localstack()%20%7B%0A%20%20%20%20%20%20%20%20shared_config%20%3D%20shared_config.endpoint_url(LOCALSTACK_ENDPOINT)%3B%0A%20%20%20%20%7D%3B).

Hence this implementation is under the tests flag.
What do you propose as a solution?, 
Do you think changing prod code as how tests are written (adding endpoint url on prod just because it's available on tests) is a valid way to go ?

crates/orchestrator/src/routes.rs
heemankv: This is a valid concern,
While the tests for `Job_controller` were being written it was realised that we are not using it anymore and hence it can be deleted, we agree, ideally this is to be done on a separate PR.

Proposed Solution :
- Have stricter decouples of PR from next time.
- Keep the commit with this PR.
- Mention removal of `job_controller` in the `CHANGELOG`.

crates/orchestrator/src/tests/database/mod.rs
heemankv: Valid,
We had used `fixture` in the initial stages of tests but we soon realised that we require more customisability like mocking only `da_client` while writing tests, hence we went with the [Builder Design Pattern](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html) to achieve the same.

Moreover : 
- `drop_database` is an async fn, we didn't find this to be a valid enough reason to implement `Drop` on Config.

Proposed Solution : 
- we add `drop_database` as a part of `TestConfigBuilder.build()` so that we drop the database once before each test-case runs, which gives same result as dropping database after each test-case.

WDYT ?

crates/orchestrator/src/config.rs
heemankv: Valid, `unwrap_or_else` can readily be used for `prover_client` for better code pattern.

Upon further inspection : 
- settlement_client
- storage_client
- da_client
- database
are built using `async build functions` which makes it tricky to create using `unwrap_or_else`.

We can write something like : 

```rust
use tokio::runtime::Handle;

// ...

self.settlement_client = Some(self.settlement_client.unwrap_or_else(|| {
    Handle::current().block_on(build_settlement_client(&settings_provider))
}));

```
But then we'll have unnecessary runtime overhead.

### Proposed Solution : 
 - change `sync build functions` to use `unwrap_or_else`.
 - handle `async build functions` in separate assignment using `await`.
 
 WDYT ?

crates/orchestrator/src/routes.rs
odesenfans: You can keep it in this PR, but mention it in the changelog and in the PR description.

crates/orchestrator/src/tests/database/mod.rs
odesenfans: Several things to say here.

1. Regarding when to drop the DB, I believe it's best to drop it when the test starts and not after it finishes. This way, you can inspect the DB to understand what went wrong.
2. Fixture vs manual call: While you're going to have multiple test configs, you're only going to use a few. So defining a few fixtures with the right configuration options on the test builder that return the test config is probably the right thing to do. This enables finer-grained control on when to reset the DB (once for a group of tests? For each test?) depending on what you want.
3. We didn't talk about test concurrency here, but if you let all your tests execute in parallel you have a strong risk of dropping the DB in the middle of another test. Do you have a solution for this?

crates/orchestrator/src/database/mongodb/mod.rs
```diff
@@ -40,6 +40,13 @@ impl MongoDb {
         MongoDb { client }
     }
 
+    /// Mongodb client uses Arc internally, reducing the cost of clone.
+    /// Directly using clone is not recommended for libraries not using Arc internally.
+    /// Dev might want to pass an Arc manually.
```
odesenfans, line 45: Not sure what "Dev might want to pass an Arc manually." means.

crates/orchestrator/src/data_storage/aws_s3/config.rs
odesenfans: But specifying the endpoint is still a possibility on AWS, right? I'm not in favor of "modifying production code", but here it's just that you have 2 different ways of connecting: either you specify the region, bucket key etc or you provide the endpoint directly. It turns out that localstack only supports the endpoint method, sure. 

In effect you have two usable connection solutions => you could use an enumeration to support both features, for example giving priority to the endpoint method if someone wants to hardcode it for some reason and then support the default method as a backup. This way, your code is the same for testing vs prod and you support two methods to connect to AWS. WDYT?

ocdbytes: Hey @odesenfans 
I agree with your implementation and it is more cleaner. I will implement this and push the code.

crates/orchestrator/src/routes.rs
ocdbytes: sure

crates/orchestrator/src/tests/database/mod.rs
ocdbytes: 1. Yes so we will do that in the beginning of the tests only.
2. Will have to think about that can you give me an example on this if possible ?
3. We were also discussing about this few days back. We will have to come up on this solution but currently we don't have one yet. We were thinking to make collections test wise or databases for different tests something like that.

crates/orchestrator/src/database/mongodb/mod.rs
ocdbytes: will remove this comment : "Dev might want to pass an Arc manually.". It may create confusion for someone.

crates/orchestrator/src/tests/database/mod.rs
apoorvsadana: To add on (3), the major blocker for concurrency would be DB or AWS localstack. We haven't finalized the implementation yet but it shouldn't be difficult to add on the current implementation. As @ocdbytes mentioned, we were looking at mostly using different databases for each test case and different localstack accounts. I think the changes should be localized to the builder itself but the idea was to focus on this more when we see test cases taking >15mins or so. wdyt?

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -69,7 +69,7 @@ impl EthereumSettlementClient {
             ProviderBuilder::new().with_recommended_fillers().wallet(wallet.clone()).on_http(settlement_cfg.rpc_url),
         );
         let core_contract_client = StarknetValidityContractClient::new(
-            Address::from_slice(settlement_cfg.core_contract_address.as_bytes()).0.into(),
+            Address::from_str(&settlement_cfg.core_contract_address).unwrap().0.into(),
```
EvolveArt, line 72: use expect here

crates/orchestrator/src/tests/database/mod.rs
```diff
@@ -1,7 +1,61 @@
...
+}
+
+/// Tests for `create_job` operation in database trait.
+/// Creates 3 jobs and asserts them.
+#[rstest]
+#[tokio::test]
+async fn test_database_create_job() -> color_eyre::Result<()> {
+    TestConfigBuilder::new().build().await;
+
+    let config = config().await;
```
EvolveArt, line 21: you can probably have a fixture to get the config, also add the #[awt] macro [attribute](https://docs.rs/rstest/latest/rstest/attr.rstest.html#async) 
```suggestion
   #[awt]
   async fn test_database_create_job(#[future] config ) -> color_eyre::Result<()> {
```

crates/orchestrator/src/tests/config.rs
```diff
@@ -0,0 +1,126 @@
...
+                    .build_test_bucket(&get_env_var_or_panic("AWS_S3_BUCKET_NAME"))
+                    .await
+                    .unwrap(),
+                _ => panic!("Unsupported Storage Client"),
+            }
+        }
+
+        drop_database().await.unwrap();
+
+        // return config and server as tuple
```
EvolveArt, line 106: comment invalid anymore?

crates/orchestrator/src/data_storage/aws_s3/mod.rs
```diff
@@ -22,25 +22,56 @@ pub struct AWSS3 {
...
+        let (config_builder, config) = match config {
+            AWSS3ConfigType::WithoutEndpoint(config) => {
+                let credentials = Credentials::new(
+                    config.s3_key_id.clone(),
+                    config.s3_key_secret.clone(),
+                    None,
+                    None,
+                    "loaded_from_custom_env",
+                );
+                let region = Region::new(config.s3_bucket_region.clone().to_string());
```
EvolveArt, line 35: duplicate code, could be moved to its own internal function `get_credentials_and_region_from_config(config: &Config)`

EvolveArt (commented):
smol review

crates/orchestrator/src/tests/config.rs
ocdbytes: removed

crates/orchestrator/src/data_storage/aws_s3/mod.rs
ocdbytes: Added the function.

crates/orchestrator/src/tests/database/mod.rs
ocdbytes: Added

odesenfans (approved):

Issue #55, opened 2024-07-24
Mark failed jobs in database.


heemankv:
Current system doesn't mark failed jobs as `Failure` status, we need to implement the same.
There are few ways to go about this : 

### Don't bubble up the error, update state and move on
    - Pro : DB is up-to-date.
    - Con : Can't implement retry logic.

### Update then bubble up
    - Pro : retries then update state.
    - Con : too complex, not for MVP.

### Dead Letter Queue 
    - Pro : easy to implement.
    - Con : too many queues.

PR #56, opened 2024-07-24
update: added termination queue

heemankv:
This PR resolves Issue #55 .

- Implemented Dead Letter Queue handler.
- Adds `last_job_status` to job metadata.
- Runs and Tests needs to spin up appropriate queues.
- Test for `handle_job_failure`.

### DIFF
--- a/crates/orchestrator/src/jobs/mod.rs
+++ b/crates/orchestrator/src/jobs/mod.rs
@@ -181,8 +181,13 @@ pub async fn handle_job_failure(id: Uuid) -> Result<()> {
     let mut job = get_job(id).await?.clone();
     let mut metadata = job.metadata.clone();
 
-    if job.status.clone() == JobStatus::Completed {
+    if job.status == JobStatus::Completed {
         return Err(eyre!("Invalid state exists on DL queue: {}", job.status.to_string()));
+    }
+    // We assume that a Failure status wil only show up if the message is sent twice from a queue
+    // Can return silently because it's already been processed.
+    else if job.status == JobStatus::Failed {
+        return Ok(());
     }
 
     metadata.insert("last_job_status".to_string(), job.status.to_string());
--- a/crates/orchestrator/src/jobs/mod.rs
+++ b/crates/orchestrator/src/jobs/mod.rs
@@ -174,13 +174,16 @@ pub async fn verify_job(id: Uuid) -> Result<()> {
 }
 
 /// Terminates the job and updates the status of the job in the DB.
-pub async fn terminate_job(id: Uuid) -> Result<()> {
+pub async fn handle_job_failure(id: Uuid) -> Result<()> {
     let config = config().await;
-    let job = get_job(id).await?;
+
+    let mut job = get_job(id).await?.clone();
     let mut metadata = job.metadata.clone();
     metadata.insert("last_job_status".to_string(), job.status.to_string());
-    config.database().update_metadata(&job, metadata).await?;
-    config.database().update_job_status(&job, JobStatus::Failed).await?;
+    job.metadata = metadata;
+    job.status = JobStatus::Failed;
+
+    config.database().update_job(&job).await?;
 
     Ok(())
 }
--- a/crates/orchestrator/src/jobs/mod.rs
+++ b/crates/orchestrator/src/jobs/mod.rs
@@ -174,11 +174,17 @@ pub async fn verify_job(id: Uuid) -> Result<()> {
 }
 
 /// Terminates the job and updates the status of the job in the DB.
+/// Throws error if the job status `Completed` is existing on DL queue.
 pub async fn handle_job_failure(id: Uuid) -> Result<()> {
     let config = config().await;
 
     let mut job = get_job(id).await?.clone();
     let mut metadata = job.metadata.clone();
+
+    if job.status.clone() == JobStatus::Completed {
+        return Err(eyre!("Invalid state exists on DL queue: {}", job.status.to_string()));
+    }
+
     metadata.insert("last_job_status".to_string(), job.status.to_string());
     job.metadata = metadata;
     job.status = JobStatus::Failed;
--- a/crates/orchestrator/src/jobs/mod.rs
+++ b/crates/orchestrator/src/jobs/mod.rs
@@ -188,7 +188,8 @@ pub async fn handle_job_failure(id: Uuid) -> Result<()> {
     let mut metadata = job.metadata.clone();
 
     if job.status == JobStatus::Completed {
-        return Err(eyre!("Invalid state exists on DL queue: {}", job.status.to_string()));
+        log::error!("Invalid state exists on DL queue: {}", job.status.to_string());
+        return Ok(());
     }
     // We assume that a Failure status wil only show up if the message is sent twice from a queue
     // Can return silently because it's already been processed.
--- a/crates/orchestrator/src/jobs/mod.rs
+++ b/crates/orchestrator/src/jobs/mod.rs
@@ -173,6 +173,18 @@ pub async fn verify_job(id: Uuid) -> Result<()> {
     Ok(())
 }
 
+/// Terminates the job and updates the status of the job in the DB.
+pub async fn terminate_job(id: Uuid) -> Result<()> {
+    let config = config().await;
+    let job = get_job(id).await?;
+    let mut metadata = job.metadata.clone();
+    metadata.insert("last_job_status".to_string(), job.status.to_string());
+    config.database().update_metadata(&job, metadata).await?;
+    config.database().update_job_status(&job, JobStatus::Failed).await?;
+
+    Ok(())
+}
+
 fn get_job_handler(job_type: &JobType) -> Box<dyn Job> {
     match job_type {
         JobType::DataSubmission => Box::new(da_job::DaJob),
--- a/crates/orchestrator/src/queue/job_queue.rs
+++ b/crates/orchestrator/src/queue/job_queue.rs
@@ -10,10 +10,11 @@ use tracing::log;
 use uuid::Uuid;
 
 use crate::config::config;
-use crate::jobs::{process_job, verify_job};
+use crate::jobs::{process_job, terminate_job, verify_job};
 
-pub const JOB_PROCESSING_QUEUE: &str = "madara_orchestrator_job_processing_queue";
-pub const JOB_VERIFICATION_QUEUE: &str = "madara_orchestrator_job_verification_queue";
+const JOB_PROCESSING_QUEUE: &str = "madara_orchestrator_job_processing_queue";
+const JOB_VERIFICATION_QUEUE: &str = "madara_orchestrator_job_verification_queue";
+const JOB_TERMINATION_QUEUE: &str = "madara_orchestrator_job_termination_queue";
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct JobQueueMessage {
@@ -84,6 +85,15 @@ pub async fn init_consumers() -> Result<()> {
             match consume_job_from_queue(JOB_VERIFICATION_QUEUE.to_string(), verify_job).await {
                 Ok(_) => {}
                 Err(e) => log::error!("Failed to consume from queue {:?}. Error: {:?}", JOB_VERIFICATION_QUEUE, e),
+            }
+            sleep(Duration::from_secs(1)).await;
+        }
+    });
+    tokio::spawn(async move {
+        loop {
+            match consume_job_from_queue(JOB_TERMINATION_QUEUE.to_string(), terminate_job).await {
+                Ok(_) => {}
+                Err(e) => log::error!("Failed to consume from queue {:?}. Error: {:?}", JOB_TERMINATION_QUEUE, e),
             }
             sleep(Duration::from_secs(1)).await;
         }
--- a/crates/orchestrator/src/queue/job_queue.rs
+++ b/crates/orchestrator/src/queue/job_queue.rs
@@ -12,10 +12,10 @@ use uuid::Uuid;
 use crate::config::config;
 use crate::jobs::{handle_job_failure, process_job, verify_job};
 
-const JOB_PROCESSING_QUEUE: &str = "madara_orchestrator_job_processing_queue";
-const JOB_VERIFICATION_QUEUE: &str = "madara_orchestrator_job_verification_queue";
+pub const JOB_PROCESSING_QUEUE: &str = "madara_orchestrator_job_processing_queue";
+pub const JOB_VERIFICATION_QUEUE: &str = "madara_orchestrator_job_verification_queue";
 // Below is the Data Letter Queue for the the above two jobs.
-const JOB_HANDLE_FAILURE_QUEUE: &str = "madara_orchestrator_job_handle_failure_queue";
+pub const JOB_HANDLE_FAILURE_QUEUE: &str = "madara_orchestrator_job_handle_failure_queue";
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct JobQueueMessage {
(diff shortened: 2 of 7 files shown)


crates/orchestrator/src/jobs/mod.rs
```diff
@@ -173,6 +173,18 @@ pub async fn verify_job(id: Uuid) -> Result<()> {
...
 }
 
+/// Terminates the job and updates the status of the job in the DB.
+pub async fn terminate_job(id: Uuid) -> Result<()> {
+    let config = config().await;
+    let job = get_job(id).await?;
+    let mut metadata = job.metadata.clone();
+    metadata.insert("last_job_status".to_string(), job.status.to_string());
+    config.database().update_metadata(&job, metadata).await?;
+    config.database().update_job_status(&job, JobStatus::Failed).await?;
```
apoorvsadana, line 183: can we just do `update_job`?

crates/orchestrator/src/jobs/mod.rs
```diff
@@ -173,6 +173,18 @@ pub async fn verify_job(id: Uuid) -> Result<()> {
     Ok(())
 }
 
+/// Terminates the job and updates the status of the job in the DB.
+pub async fn terminate_job(id: Uuid) -> Result<()> {
```
apoorvsadana, line 177:
```suggestion
pub async fn handle_job_failure(id: Uuid) -> Result<()> {
```

apoorvsadana: terminate sounds like stopping a job in the middle, wdyt?

crates/orchestrator/src/queue/job_queue.rs
```diff
@@ -10,10 +10,11 @@ use tracing::log;
 use uuid::Uuid;
 
 use crate::config::config;
-use crate::jobs::{process_job, verify_job};
+use crate::jobs::{process_job, terminate_job, verify_job};
 
 const JOB_PROCESSING_QUEUE: &str = "madara_orchestrator_job_processing_queue";
 const JOB_VERIFICATION_QUEUE: &str = "madara_orchestrator_job_verification_queue";
+const JOB_TERMINATION_QUEUE: &str = "madara_orchestrator_job_termination_queue";
```
apoorvsadana, line 17:
```suggestion
const JOB_HANDLE_FAILURE_QUEUE: &str = "madara_orchestrator_job_handle_failure_queue";
```

crates/orchestrator/src/queue/job_queue.rs
```diff
@@ -88,6 +89,15 @@ pub async fn init_consumers() -> Result<()> {
             sleep(Duration::from_secs(1)).await;
         }
     });
+    tokio::spawn(async move {
```
apoorvsadana, line 92: did you see if we can generalise this with a `macro!`

crates/orchestrator/src/queue/job_queue.rs
apoorvsadana, line 17: let's add a comment here that this should be a DLQ for the two queues above.

crates/orchestrator/src/jobs/mod.rs
heemankv: No issues, changing to handle_job_failure

crates/orchestrator/src/jobs/mod.rs
heemankv: Valid, updated.

apoorvsadana (approved):

crates/orchestrator/src/jobs/mod.rs
```diff
@@ -173,6 +173,21 @@ pub async fn verify_job(id: Uuid) -> Result<()> {
     Ok(())
 }
 
+/// Terminates the job and updates the status of the job in the DB.
+pub async fn handle_job_failure(id: Uuid) -> Result<()> {
```
apoorvsadana, line 177: let's add a test case after the fixture PR is merged.

heemankv: Added.

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -20,3 +25,103 @@ async fn create_job_fails_job_already_exists() {
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "Created" => Ok(JobStatus::Created),
+            "LockedForProcessing" => Ok(JobStatus::LockedForProcessing),
+            "PendingVerification" => Ok(JobStatus::PendingVerification),
+            "Completed" => Ok(JobStatus::Completed),
+            "VerificationTimeout" => Ok(JobStatus::VerificationTimeout),
+            "Failed" => Ok(JobStatus::Failed),
+            s if s.starts_with("VerificationFailed(") && s.ends_with(')') => {
```
apoorvsadana, line 42: didn't we remove the reason in `VerificationFailed`?

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -20,3 +25,103 @@ async fn create_job_fails_job_already_exists() {
...
+#[case("DataSubmission", "Failed")]
+#[case("StateTransition", "Completed")]
+#[case("ProofCreation", "VerificationTimeout")]
+#[case("DataSubmission", "VerificationFailed()")]
+#[tokio::test]
+async fn test_handle_job_failure(#[case] job_type: JobType, #[case] job_status: JobStatus) -> color_eyre::Result<()> {
+    use color_eyre::eyre::eyre;
+
+    TestConfigBuilder::new().build().await;
+    dotenvy::from_filename("../.env.test")?;
```
apoorvsadana, line 80: doesn't this happen in the builder already?

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -20,3 +25,103 @@ async fn create_job_fails_job_already_exists() {
...
+        job.metadata = metadata;
+    }
+
+    // feeding the job to DB
+    database_client.create_job(job.clone()).await.unwrap();
+
+    // calling handle_job_failure
+    let response = handle_job_failure(job_id).await;
+
+    match response {
```
apoorvsadana, line 104: i think we should just write different test cases, there isn't a lot of common code here

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -20,3 +25,103 @@ async fn create_job_fails_job_already_exists() {
...
+
+#[rstest]
+#[case("DataSubmission", "Completed")] // code should panic here, how can completed move to dl queue ?
+#[case("SnosRun", "PendingVerification")]
+#[case("ProofCreation", "LockedForProcessing")]
+#[case("ProofRegistration", "Created")]
+#[case("DataSubmission", "Failed")]
+#[case("StateTransition", "Completed")]
+#[case("ProofCreation", "VerificationTimeout")]
+#[case("DataSubmission", "VerificationFailed()")]
```
apoorvsadana, line 74: we can just do one working case

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -20,3 +25,103 @@ async fn create_job_fails_job_already_exists() {
...
+    match response {
+        Ok(()) => {
+            // check job in db
+            let job = config.database().get_job_by_id(job_id).await?;
+
+            if let Some(job_item) = job {
+                // check if job status is Failure
+                assert_eq!(job_item.status, JobStatus::Failed);
+                // check if job metadata has `last_job_status`
+                assert_ne!(None, job_item.metadata.get("last_job_status"));
```
apoorvsadana, line 113: why not equal?

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -20,3 +25,103 @@ async fn create_job_fails_job_already_exists() {
 async fn create_job_fails_works_new_job() {
     // TODO
 }
+
+use std::str::FromStr;
+
+impl FromStr for JobStatus {
```
apoorvsadana, line 31: what do we need this and the below one for

crates/orchestrator/src/jobs/mod.rs
```diff
@@ -173,6 +173,32 @@ pub async fn verify_job(id: Uuid) -> Result<()> {
...
+/// Terminates the job and updates the status of the job in the DB.
+/// Throws error if the job status `Completed` is existing on DL queue.
+pub async fn handle_job_failure(id: Uuid) -> Result<()> {
+    let config = config().await;
+
+    let mut job = get_job(id).await?.clone();
+    let mut metadata = job.metadata.clone();
+
+    if job.status == JobStatus::Completed {
+        return Err(eyre!("Invalid state exists on DL queue: {}", job.status.to_string()));
```
apoorvsadana, line 185: let's log the error and add a TODO to send an alert. we don't really want to retry from SQS here

crates/orchestrator/src/tests/jobs/mod.rs
heemankv: Yes, 
you are correct.
It was removed as a part of #52, rebased/removed.

crates/orchestrator/src/tests/jobs/mod.rs
heemankv: Valid, removed.

crates/orchestrator/src/tests/jobs/mod.rs
heemankv: These `impl` help `rstest` cases to convert from `String` input to `JobType/JobStatus` output to be used as function arguments.

crates/orchestrator/src/jobs/mod.rs
heemankv: Valid, removed.

crates/orchestrator/src/tests/jobs/mod.rs
heemankv: We ensure that if the `jobStatus` is `Failed` then the `job.metadata` files should have the field `last_job_status`.

heemankv: On second thought, it makes more sense to use :
`assert!(job_item.metadata.get("last_job_status").is_some());`
I have applied the same.

crates/orchestrator/src/tests/jobs/mod.rs
heemankv: Valid, removed.

crates/orchestrator/src/tests/jobs/mod.rs
heemankv: Valid, seperated the test cases.

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -12,11 +17,120 @@ pub mod state_update_job;
...
+            panic!("Test case should have passed: {} ", err);
+        }
+    }
+}
+
+#[rstest]
+// code should panic here, how can completed move to dl queue ?
+#[case("DataSubmission")]
+#[tokio::test]
+async fn handle_job_failure__job_status_completed_works(#[case] job_type: JobType) {
```
apoorvsadana, line 112:
```suggestion
async fn handle_job_failure_with_completed_job_status_works(#[case] job_type: JobType) {
```

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -12,11 +17,120 @@ pub mod state_update_job;
...
+            }
+        }
+        Err(err) => {
+            panic!("Test case should have passed: {} ", err);
+        }
+    }
+}
+
+#[rstest]
+// code should panic here, how can completed move to dl queue ?
```
apoorvsadana, line 109: is this comment resolvd?

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -12,11 +17,120 @@ pub mod state_update_job;
...
+    database_client.create_job(job.clone()).await.unwrap();
+
+    // calling handle_job_failure
+    handle_job_failure(job_id).await.expect("Test call to handle_job_failure should have passed.");
+
+    // The completed job status on db is untouched.
+    let job_fetched_result = config.database().get_job_by_id(job_id).await.expect("Unable to fetch Job Data");
+
+    if let Some(job_fetched) = job_fetched_result {
+        assert_eq!(job_fetched.status, JobStatus::Completed);
```
apoorvsadana, line 134: we can assert entire jobs are equal

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -12,11 +17,120 @@ pub mod state_update_job;
...
+            _ => Err(format!("Invalid job type: {}", s)),
+        }
+    }
+}
+
+#[rstest]
+#[case("SnosRun", "PendingVerification")]
+#[case("DataSubmission", "Failed")]
+#[tokio::test]
+async fn handle_job_failure_job_status_typical_works(#[case] job_type: JobType, #[case] job_status: JobStatus) {
```
apoorvsadana, line 65:
```suggestion
async fn handle_job_failure_with_correct_job_status_works(#[case] job_type: JobType, #[case] job_status: JobStatus) {
```

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -12,11 +17,120 @@ pub mod state_update_job;
...
+    match response {
+        Ok(()) => {
+            // check job in db
+            let job = config.database().get_job_by_id(job_id).await.expect("Unable to fetch Job Data");
+
+            if let Some(job_item) = job {
+                // check if job status is Failure
+                assert_eq!(job_item.status, JobStatus::Failed);
+                // check if job metadata has `last_job_status`
+                assert!(job_item.metadata.get("last_job_status").is_some());
```
apoorvsadana, line 98: we should check the actual job status

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -12,11 +17,120 @@ pub mod state_update_job;
...
+        job.metadata = metadata;
+    }
+
+    // feeding the job to DB
+    database_client.create_job(job.clone()).await.unwrap();
+
+    // calling handle_job_failure
+    let response = handle_job_failure(job_id).await;
+
+    match response {
```
apoorvsadana, line 89: let's use `assert_matches!`? i think we can match only some fields using something like this (taken from claude)

```rust
let person = Person {
        name: String::from("Alice"),
        age: 30,
        city: String::from("New York"),
    };

    assert_matches!(person, Person { name: ref n, age: 30, .. } => {
        assert_eq!(n, "Alice");
    });
```

crates/orchestrator/src/tests/jobs/mod.rs
apoorvsadana: can we not use something like this - https://docs.rs/rstest/latest/rstest/attr.rstest.html#local-lifetime-and-by_ref-attribute

crates/orchestrator/src/tests/jobs/mod.rs
heemankv: Valid, Applied

crates/orchestrator/src/tests/jobs/mod.rs
heemankv: Yes, we are assuming queue error leading to this error.

crates/orchestrator/src/tests/jobs/mod.rs
heemankv: Valid, I am assuming this is possible because `JobItem` and every other struct that is used in it like `JobType` and `JobStatus` implement `Partial_Eq` trait.

apoorvsadana:
Closing in favor of #73

Issue #65, opened 2024-07-31
create custom Job Errors


heemankv:
### Problem Statement
The codebase has extensive use of `color_eyre` for handling errors.
We would want to move away from this and use `thisError` for all the Jobs,

### Expected Implementation
For all jobs inside `orchestrator/src/jobs` we want to `not return` `color_eyre::Result`. 
Accordingly we want to define errors with `thisError`.

PR #66, opened 2024-07-31
Update/job error using this error

heemankv:
# This PR solves #65.

- This PR introduces `thiserror` for clearer description and syntax of throwing errors.
- This PR scopes for `Jobs` only!
- Further implementation to overall codebase are to be handled incrementally.
- This PR depends on PR #61 and PR #52 (only after they are merged should this branch be merged to main.)
- Implementation also removes `log::debug!` from various places for which `thiserror` error handling is introduced.
- Only Errors specific to the `Job` are handled, general errors are thrown as `Other Errors`.

### DIFF
--- a/crates/orchestrator/src/jobs/da_job/mod.rs
+++ b/crates/orchestrator/src/jobs/da_job/mod.rs
@@ -38,6 +38,21 @@ lazy_static! {
     pub static ref BLOB_LEN: usize = 4096;
 }
 
+#[derive(Error, Debug)]
+pub enum DaError {
+    #[error("Cannot process block {block_no:?} for job id {job_id:?} as it's still in pending state.")]
+    BlockPending { block_no: u64, job_id: Uuid },
+
+    #[error("Blob size must be at least 32 bytes to accommodate a single FieldElement/BigUint, but was {blob_size:?}")]
+    InsufficientBlobSize { blob_size: u64 },
+
+    #[error("Exceeded the maximum number of blobs per transaction: allowed {max_blob_per_txn:?}, found {current_blob_length:?} for block {block_no:?} and job id {job_id:?}")]
+    MaxBlobsLimitExceeded { max_blob_per_txn: u64, current_blob_length: u64, block_no: u64, job_id: Uuid },
+
+    #[error("Other error: {0}")]
+    Other(#[from] color_eyre::eyre::Error),
+}
+
 pub struct DaJob;
 
 #[async_trait]
@@ -60,15 +75,13 @@ impl Job for DaJob {
     }
 
     async fn process_job(&self, config: &Config, job: &mut JobItem) -> Result<String, JobError> {
-        let block_no = job.internal_id.parse::<u64>()
-        .wrap_err("Failed to parse u64".to_string())?;
+        let block_no = job.internal_id.parse::<u64>().wrap_err("Failed to parse u64".to_string())?;
 
         let state_update = config
             .starknet_client()
             .get_state_update(BlockId::Number(block_no))
             .await
             .wrap_err("Failed to get state Update.".to_string())?;
-
 
         let state_update = match state_update {
             MaybePendingStateUpdate::PendingUpdate(_) => Err(DaError::BlockPending { block_no, job_id: job.id })?,
@@ -87,10 +100,8 @@ impl Job for DaJob {
 
         // converting BigUints to Vec<u8>, one Vec<u8> represents one blob data
         let blob_array = data_to_blobs(max_bytes_per_blob, transformed_data)?;
-        let current_blob_length: u64 = blob_array
-            .len()
-            .try_into()
-            .wrap_err("Unable to convert the blob length into u64 format.".to_string())?;
+        let current_blob_length: u64 =
+            blob_array.len().try_into().wrap_err("Unable to convert the blob length into u64 format.".to_string())?;
 
         // there is a limit on number of blobs per txn, checking that here
         if current_blob_length > max_blob_per_txn {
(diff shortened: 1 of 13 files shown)


crates/orchestrator/src/jobs/state_update_job/mod.rs
```diff
@@ -222,3 +243,45 @@ impl StateUpdateJob {
         job.metadata.insert(new_attempt_metadata_key, tx_hashes.join(","));
     }
 }
+
+#[derive(Error, Debug)]
+pub enum StateUpdateError {
```
apoorvsadana, line 248: let's move all errors to top of file

crates/orchestrator/src/queue/job_queue.rs
```diff
@@ -20,20 +20,20 @@ pub struct JobQueueMessage {
...
 }
 
-pub async fn add_job_to_verification_queue(id: Uuid, delay: Duration) -> Result<()> {
+pub async fn add_job_to_verification_queue(id: Uuid, delay: Duration) -> EyreResult<()> {
     log::info!("Adding job with id {:?} to verification queue", id);
     add_job_to_queue(id, JOB_VERIFICATION_QUEUE.to_string(), Some(delay)).await
 }
 
-pub async fn consume_job_from_queue<F, Fut>(queue: String, handler: F) -> Result<()>
+pub async fn consume_job_from_queue<F, Fut>(queue: String, handler: F) -> EyreResult<()>
```
apoorvsadana, line 33: should we add custom errors here as well? `FailedToConsumeFromQueue`, `FailedToHandleJob`

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -61,20 +59,17 @@ impl Job for DaJob {
         })
     }
 
-    async fn process_job(&self, config: &Config, job: &mut JobItem) -> Result<String> {
-        let block_no = job.internal_id.parse::<u64>()?;
+    async fn process_job(&self, config: &Config, job: &mut JobItem) -> Result<String, JobError> {
+        let block_no = job.internal_id.parse::<u64>().wrap_err_with(|| "Failed to parse u64".to_string())?;
```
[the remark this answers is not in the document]
heemankv: Valid, applied.

crates/orchestrator/src/jobs/state_update_job/mod.rs
heemankv: unaware of this convention, Moved to top.

crates/orchestrator/src/queue/job_queue.rs
heemankv:
```rust
pub async fn init_consumers() -> Result<(), JobError> {
    // TODO: figure out a way to generalize this
    tokio::spawn(async move {
        loop {
            match consume_job_from_queue(JOB_PROCESSING_QUEUE.to_string(), process_job).await {
                Ok(_) => {}
                Err(e) => log::error!("Failed to consume from queue {:?}. Error: {:?}", JOB_PROCESSING_QUEUE, e),
            }
        }
    });
```
### My thought process :
I checked the usage of `consume_job_from_queue` and it's used in `init_consumers`,
while matching the error we are only logging the error and moving ahead.
This code indirectly suggested that throwing custom error for `consume_job_from_queue` is not as important as others as it is no where stopping the code.

Though this is not a fact based decision, do you see any perks in creating custom errors here ? 
@apoorvsadana

apoorvsadana: But the logging part is mostly true for other errors as well right? I think we're logging it in almost every case for now but as `lib`, it makes it generalised so that others could handle it if they like. One example is in test cases, we check for `QueueError` to have `NoData`

heemankv:
```rust
 match job_message {
        Some(job_message) => {
            log::info!("Handling job with id {:?} for queue {:?}", job_message.id, queue);
            match handler(job_message.id).await {
                Ok(_) => delivery.ack().await.map_err(|(e, _)| e)?,
                Err(e) => {
                    log::error!("Failed to handle job with id {:?}. Error: {:?}", job_message.id, e);

                    // if the queue as a retry logic at the source, it will be attempted
                    // after the nack
                    delivery.nack().await.map_err(|(e, _)| e)?;
                }
            };
        }
        None => return Ok(()),
    };
```

Valid, in the above code the `handler` is the general `process_job` function, which is also being logged.
Makes sense, adding the Errors.

heemankv: Added the needed Error cases 
Please check @apoorvsadana.

crates/orchestrator/src/jobs/mod.rs
```diff
@@ -183,38 +211,42 @@ fn get_job_handler(job_type: &JobType) -> Box<dyn Job> {
...
 
-fn increment_key_in_metadata(metadata: &HashMap<String, String>, key: &str) -> Result<HashMap<String, String>> {
+fn increment_key_in_metadata(
+    metadata: &HashMap<String, String>,
+    key: &str,
+) -> Result<HashMap<String, String>, JobError> {
     let mut new_metadata = metadata.clone();
     let attempt = get_u64_from_metadata(metadata, key)?;
     let incremented_value = attempt.checked_add(1);
     if incremented_value.is_none() {
```
cchudant, line 230:
```rs
let Some(incremented_value) = incremented_value else {
  return Err(JobError::KeyOutOfBounds { key: key.to_string() }) }
};
``` to avoid the unwrap just after

cchudant (approved):
cool, looks good to me :)

crates/orchestrator/src/jobs/mod.rs
cchudant: or in this case just, `incremented_value.ok_or_else(|| JobError::KeyOutOfBounds { key: key.to_string() })`

heemankv: Good catch !
Applied.

PR #72, opened 2024-08-06
Test : state update job

ocdbytes:
- Added tests for state update job.


crates/orchestrator/src/tests/jobs/state_update_job/mod.rs
```diff
@@ -21,16 +21,89 @@ use crate::jobs::{
...
 }
 
 pub const X_0_FILE_NAME: &str = "x_0.txt";
 
+// ================= Exhaustive tests (with minimum mock) =================
+
+#[rstest]
+#[should_panic(expected = "Could not find current attempt number.")]
+#[tokio::test]
+async fn test_process_job_attempt_no_not_present_fails() {
```
apoorvsadana, line 43:
```suggestion
async fn test_process_job_attempt_not_present_fails() {
```

crates/orchestrator/src/tests/jobs/state_update_job/mod.rs
```diff
@@ -21,16 +21,89 @@ use crate::jobs::{
...
+    // Building new config with mocked settlement client
+    TestConfigBuilder::new().mock_settlement_client(Box::new(settlement_client)).build().await;
+
+    // setting last failed block number as 651053.
+    // setting blocks yet to process as 651054 and 651055.
+    // here total blocks to process will be 3.
+    let mut metadata: HashMap<String, String> = HashMap::new();
+    metadata.insert(JOB_PROCESS_ATTEMPT_METADATA_KEY.to_string(), "0".to_string());
+    metadata.insert(JOB_METADATA_STATE_UPDATE_LAST_FAILED_BLOCK_NO.to_string(), "651053".to_string());
+    metadata.insert(JOB_METADATA_STATE_UPDATE_BLOCKS_TO_SETTLE_KEY.to_string(), "651053,651054,651055".to_string());
```
apoorvsadana, line 92: can we do cases for this

1. no failed block no present
2. failed block present (in the middle so that we can check that processed ones weren't processed again)

ocdbytes: Added the test cases.

apoorvsadana (approved):

crates/orchestrator/src/tests/jobs/state_update_job/mod.rs
```diff
@@ -21,16 +22,98 @@ use crate::jobs::{
...
 lazy_static! {
     pub static ref CURRENT_PATH: PathBuf = std::env::current_dir().unwrap();
 }
 
 pub const X_0_FILE_NAME: &str = "x_0.txt";
 
+// ================= Exhaustive tests (with minimum mock) =================
+
+#[rstest]
+#[should_panic(expected = "Could not find current attempt number.")]
```
odesenfans, line 42: Should panic? Why?

crates/orchestrator/src/tests/jobs/state_update_job/mod.rs
```diff
@@ -21,16 +22,98 @@ use crate::jobs::{
...
+
+    // Storing `blob_data` and `snos_output` in storage client
+    store_data_in_storage_client_for_s3(block_numbers.clone()).await;
+
+    // Building a temp config that will be used by `fetch_blob_data_for_block` and `fetch_snos_for_block`
+    // functions while fetching the blob data from storage client.
+    TestConfigBuilder::new().build().await;
+
+    // Adding expectations for each block number to be called by settlement client.
+    for (i, _) in block_numbers.iter().enumerate().skip(processing_start_index as usize) {
```
odesenfans, line 81:
```suggestion
    for i in (processing_start_index as usize)..block_numbers.len() {
```

crates/orchestrator/src/tests/jobs/state_update_job/mod.rs
ocdbytes: Because the function should panic when attempt number is not found in job metadata.

odesenfans: Wouldn't returning an error be better?

crates/orchestrator/src/tests/jobs/state_update_job/mod.rs
odesenfans: That sounds weird. Did you try this exact syntax?

crates/orchestrator/src/tests/jobs/state_update_job/mod.rs
ocdbytes: yeah so we have a PR that is pending and includes the implementation of thiserror library. So in that case I will update the test to return that particular error. So for now this is added as panic!

crates/orchestrator/src/tests/jobs/state_update_job/mod.rs
ocdbytes: yeah little bit different than this. but sure I will try your syntax and update it here.

ocdbytes: For you syntax also this is the error I am getting, which is same when I used this syntax. 

```sh
error: the loop variable `i` is only used to index `block_numbers`
  --> crates/orchestrator/src/tests/jobs/state_update_job/mod.rs:79:14
   |
79 |     for i in (processing_start_index as usize)..block_numbers.len() {
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_range_loop
   = note: `-D clippy::needless-range-loop` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::needless_range_loop)]`
help: consider using an iterator
   |
79 |     for <item> in block_numbers.iter().skip((processing_start_index as usize)) {
   |         ~~~~~~    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

odesenfans: Ah yeah but then Clippy is giving you the right advice, you do not need `i` at all. Just apply the suggestion of Clippy.

odesenfans (approved):

PR #73, opened 2024-08-08
Feature: data letter queue

heemankv:
This PR resolves Issue #55 .

- Implemented Dead Letter Queue handler.
- Adds `last_job_status` to job metadata.
- Runs and Tests needs to spin up appropriate queues.
- Test for `handle_job_failure`.


apoorvsadana (commented):
Original PR was #56. Created this because of merge conflicts as explained on Slack

crates/orchestrator/src/jobs/mod.rs
```diff
@@ -186,6 +186,33 @@ pub async fn verify_job(id: Uuid) -> Result<()> {
...
+/// Logs error if the job status `Completed` is existing on DL queue.
+pub async fn handle_job_failure(id: Uuid) -> Result<()> {
+    let config = config().await;
+
+    let mut job = get_job(id).await?.clone();
+    let mut metadata = job.metadata.clone();
+
+    if job.status == JobStatus::Completed {
+        log::error!("Invalid state exists on DL queue: {}", job.status.to_string());
+        return Ok(());
```
EvolveArt, line 199: why do we fail silently here ?

crates/orchestrator/src/jobs/types.rs
```diff
@@ -99,6 +99,22 @@ pub enum JobStatus {
     VerificationTimeout,
     /// The job failed processing
     VerificationFailed,
+    /// The job failed completing
+    Failed,
+}
+
+impl fmt::Display for JobStatus {
```
EvolveArt, line 106: let's use [strum](https://docs.rs/strum/latest/strum/derive.Display.html)

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -501,3 +507,90 @@ fn build_job_item_by_type_and_status(job_type: JobType, job_status: JobStatus, i
...
+    handle_job_failure(job_id).await.expect("handle_job_failure failed to run");
+
+    let job_fetched = config.database().get_job_by_id(job_id).await.expect("Unable to fetch Job Data").unwrap();
+
+    assert_eq!(job_fetched, job_expected);
+}
+
+#[rstest]
+#[case::pending_verification(JobType::SnosRun, JobStatus::PendingVerification)]
+#[case::verification_timeout(JobType::SnosRun, JobStatus::VerificationTimeout)]
```
EvolveArt, line 541: why are the other statuses not covered ?

EvolveArt (changes requested):
We should probably add some tests for the edge cases where it would fail ? 
Also I'm correct that in this PR the DLQ is implemented but is never used by the orchestrator right ?

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -501,3 +507,90 @@ fn build_job_item_by_type_and_status(job_type: JobType, job_status: JobStatus, i
         version: 0,
     }
 }
+
+#[rstest]
+#[case(JobType::SnosRun, JobStatus::Failed)]
+#[tokio::test]
+async fn handle_job_failure_with_failed_job_status_works(#[case] job_type: JobType, #[case] job_status: JobStatus) {
+    TestConfigBuilder::new().build().await;
```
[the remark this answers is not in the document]
heemankv: As mentioned in [here](https://github.com/madara-alliance/madara-orchestrator/pull/61#discussion_r1709195363
).

`TestConfigBuilder` allows for customisation over any `external client` of our choice,
moving it to a global fixture is not feasible since all test cases have different customisation requirements.

We can make fixture for tests under same scope if they require same customised external clients.

for eg : 
All tests under `da_job` if require same config customisation, can implement a `fixture` just for themselves.
similarly for other scopes .

We can create a separate issue for this and resolve there.

crates/orchestrator/src/jobs/mod.rs
heemankv: DL-queue is supposed to handle actual failed cases.
If `JobStatus::Completed` job is pushed to DL-queue multiple times by the queuing agent,
we prefer not stopping the orchestrator rather `failing silently`.

crates/orchestrator/src/tests/jobs/mod.rs
heemankv: We initially covered all the statuses as shown [here](https://github.com/madara-alliance/madara-orchestrator/blob/8db6859f7b53e4e205e2ec0276fbb309a579197b/crates/orchestrator/src/tests/jobs/mod.rs#L545), but it felt redundant to test all.

What do you suggest @EvolveArt  ?

crates/orchestrator/src/jobs/types.rs
heemankv: Good Catch !
Implemented.

heemankv:
> We should probably add some tests for the edge cases where it would fail ?
> Also I'm correct that in this PR the DLQ is implemented but is never used by the orchestrator right ?

Please feel free to mention the edge case to run the tests on.
And yes this PR is for adding support of DLQ, after which we would setup DL-queue to be used in the future. 
@apoorvsadana feel free to add more!
Thanks

apoorvsadana:
@EvolveArt the DLQ is being used. You need to set it up on SQS/RabbitMQ etc. and when messages fail, they automatically go to the DLQ and they are moved to the failed state.

crates/orchestrator/src/tests/config.rs
```diff
@@ -127,6 +127,8 @@ impl TestConfigBuilder {
             self.storage.unwrap(),
         );
 
+        drop_database().await.unwrap();
```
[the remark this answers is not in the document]
EvolveArt: yea the question was more, why do we drop in the middle of the code

crates/orchestrator/src/tests/jobs/mod.rs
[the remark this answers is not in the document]
EvolveArt: sounds good to me

crates/orchestrator/src/tests/jobs/mod.rs
EvolveArt: well even if it feels redundant I think it's important to have them, you want to avoid having jobs at an unexpected state in the DLQ

EvolveArt (approved):
LGTM

crates/orchestrator/src/tests/jobs/mod.rs
heemankv: Valid, Implemented

crates/orchestrator/src/tests/config.rs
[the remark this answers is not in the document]
heemankv: Since this is an initiation function for all the clients, which runs before each test we can drop the database at any point in this code.

PR #76, opened 2024-08-08
Fix : fixed the worker logic and updated the tests

ocdbytes:
- Current Scenario :
    - We are taking the last successful job and then fetching the successful proving jobs.
    - Then we are creating jobs for each of the blocks and their proofs which will run parallelly.
    - This will lead to nonce issues when we are settling the transactions on ethereum.
- Required scenario :
    - We take the last successful jobs.
    - We take the successful proving jobs and collect their block numbers and then we pass these block numbers into the metadata.
    - Then we settle the transactions in one go on ethereum. (this logic is implemented in our state update job)
- Fixed state update worker logic as per the new implementation.


crates/orchestrator/src/workers/update_state.rs
```diff
@@ -45,3 +51,15 @@ impl Worker for UpdateStateWorker {
...
+
+impl UpdateStateWorker {
+    /// To parse the block numbers from the vector of jobs.
+    fn parse_job_items_into_block_number_list(job_items: Vec<JobItem>) -> String {
+        let mut block_number_string = String::from("");
+        for job in job_items {
+            block_number_string.push_str(&job.internal_id);
+            block_number_string += ",";
+        }
+        block_number_string
```
apoorvsadana, line 63:
```suggestion
        job_items.iter().map(|j| j.internal_id).collect::<Vec<String>>().join(",");
```

apoorvsadana (commented):
Can you add small description on why are we doing these changes for the external reviewer

crates/orchestrator/src/workers/update_state.rs
```diff
@@ -45,3 +51,10 @@ impl Worker for UpdateStateWorker {
         }
     }
 }
+
+impl UpdateStateWorker {
+    /// To parse the block numbers from the vector of jobs.
+    fn parse_job_items_into_block_number_list(job_items: Vec<JobItem>) -> String {
```
antiyro, line 57: Do you have some tests for this new method already?

antiyro: Would be great to cover it already

crates/orchestrator/src/tests/workers/update_state/mod.rs
```diff
@@ -78,18 +76,22 @@ async fn test_update_state_worker(
...
+            .withf(move |item| item.internal_id == *"1".to_string())
+            .returning(move |_| Ok(job_item_cloned.clone()));
     }
 
     let y: Arc<Box<dyn Job>> = Arc::new(Box::new(job_handler));
     let ctx = mock_factory::get_job_handler_context();
     // Mocking the `get_job_handler` call in create_job function.
     if last_successful_job_exists {
-        ctx.expect().times(5).with(eq(JobType::StateTransition)).returning(move |_| Arc::clone(&y));
+        ctx.expect().times(1).with(eq(JobType::StateTransition)).returning(move |_| Arc::clone(&y));
```
antiyro, line 94: could we avoid some magic numbers here?

crates/orchestrator/src/tests/workers/update_state/mod.rs
```diff
@@ -78,18 +76,22 @@ async fn test_update_state_worker(
...
         // mocking the creation of jobs
-        db_create_job_expectations_update_state_worker(
-            &mut db,
-            get_job_by_mock_id_vector(JobType::ProofCreation, JobStatus::Completed, number_of_processed_jobs as u64, 2),
-            &mut job_handler,
-        );
+        let job_item = get_job_item_mock_by_id("1".to_string(), Uuid::new_v4());
+        let job_item_cloned = job_item.clone();
+
+        job_handler.expect_create_job().times(1).returning(move |_, _, _| Ok(job_item.clone()));
```
antiyro, line 82: same as below here (?)

crates/orchestrator/src/tests/workers/update_state/mod.rs
ocdbytes: Hey sorry, I didn't understand can you please elaborate what do you mean by magic numbers ?

crates/orchestrator/src/workers/update_state.rs
ocdbytes: Sure I will add a simple test in this file itself.

crates/orchestrator/src/tests/workers/update_state/mod.rs
antiyro: by magic numbers I mean values that we should guess the meaning that could be stored in a defined constant. But I just realized we were on a testing part of the codebase so, it may not be relevent

ocdbytes: oh ok cool cool 😅

crates/orchestrator/src/workers/update_state.rs
ocdbytes: Added a simple test case

PR #81, opened 2024-08-12
Test: Ethereum Settlement Client

heemankv:
# This PR adds tests for `Ethereum Settlement Client`.
 
 ## External Dependencies : 
 - Anvil for forking and running local test chain.
 - Ethereum RPC.
 
 
## Points to note : 
- Test code blocks are included with `#[cfg(test)]` between prod code, to ensure possibility of testing, please feel free to suggest any other way.
- Two tests :  `update_state_blob_with_dummy_contract_works` & `update_state_blob_with_impersonation_works` use a  test param to make aware about impersonation or dummy contract call.
    
- `Nonce`: To ensure consistency in nonce while doing multiple `update state` transactions we extract the `nonce` fetch call from the child : `update_state_with_blobs` and manage `nonce` from `process_job`, this doesn't affect the `distributed` fashion of code since each machine will have a pre-existing lock over the jobs.   
( `update_state_with_blobs` is called by `update_state_for_block` which is called inside a loop by `process_job`)


crates/settlement-clients/ethereum/src/conversion.rs
```diff
@@ -1,12 +1,325 @@
...
+    #[case::empty(&[], U256::ZERO)]
+    fn slice_u8_to_u256_works(#[case] slice: &[u8], #[case] expected: U256) {
+        match slice_u8_to_u256(slice) {
+            Ok(response) => {
+                assert_eq!(response, expected);
+            }
+            Err(e) => {
+                panic!("{}", e);
+            }
+        }
```
apoorvsadana, line 153:
```suggestion
        assert_eq!(slice_u8_to_u256(slice).expect("slice_u8_to_u256 failed"), expected)
```

apoorvsadana: the same can be applied for stuff below. just a nitpick

crates/settlement-clients/ethereum/src/conversion.rs
```diff
@@ -1,12 +1,325 @@
...
+    }
+
+    #[rstest]
+    #[case::over(&[0xFF; 33])]
+    fn slice_u8_to_u256_panics(#[case] slice: &[u8]) {
```
apoorvsadana, line 158: I think you can add #[should_panic] on top of a specific #[case] and it only applies to that, we can combine this and the above one?

crates/settlement-clients/ethereum/src/conversion.rs
```diff
@@ -1,12 +1,325 @@
...
 use alloy::primitives::U256;
+use alloy_primitives::FixedBytes;
+use c_kzg::{Blob, KzgCommitment, KzgProof, KzgSettings};
+use color_eyre::{eyre::ContextCompat, Result as EyreResult};
+use std::fmt::Write;
 
 /// Converts a `&[Vec<u8>]` to `Vec<U256>`. Each inner slice is expected to be exactly 32 bytes long.
 /// Pads with zeros if any inner slice is shorter than 32 bytes.
-pub(crate) fn slice_slice_u8_to_vec_u256(slices: &[[u8; 32]]) -> Vec<U256> {
+pub(crate) fn slice_slice_u8_to_vec_u256(slices: &[[u8; 32]]) -> EyreResult<Vec<U256>> {
```
apoorvsadana, line 11: this function name is pretty confusing. can we do `vec_u8_32_to_vec_u256`? since it's one line, i would also question if we can just remove this function and use the line directly on the places where it's called.

crates/settlement-clients/ethereum/src/conversion.rs
```diff
@@ -1,12 +1,325 @@
...
+    fn to_hex_string_works(#[case] slice: &[u8], #[case] expected: String) {
+        let result = to_padded_hex(slice);
+        assert_eq!(result, expected);
+        assert!(expected.len() == 64);
+    }
+
+    #[rstest]
+    #[should_panic(expected = "Slice length must not exceed 32")]
+    #[case::exceeding(&[0xFF; 40], format!("{}", "ff".repeat(32)))]
+    fn to_hex_string_panics(#[case] slice: &[u8], #[case] expected: String) {
```
apoorvsadana, line 224: same as above, i think we can combine test cases

crates/settlement-clients/ethereum/src/conversion.rs
```diff
@@ -1,12 +1,325 @@
...
+        ],
+        format!("{}{}", "ff".repeat(32), "f5".repeat(32))
+    )]
+    fn vec_u8_32_to_hex_string_works(#[case] slice: Vec<[u8; 32]>, #[case] expected: String) {
+        let result = vec_u8_32_to_hex_string(slice);
+        assert_eq!(result, expected);
+    }
+
+    #[rstest]
+    #[case("20462788")]
```
apoorvsadana, line 266: let's add a small desc saying that we're matching commitments with mainnet over here

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -1,87 +1,135 @@
...
+    #[cfg(not(test))]
     pub fn with_settings(settings: &impl SettingsProvider) -> Self {
         let settlement_cfg: EthereumSettlementConfig = settings.get_settings(SETTLEMENT_SETTINGS_NAME).unwrap();
 
         let private_key = get_env_var_or_panic(ENV_PRIVATE_KEY);
         let signer: PrivateKeySigner = private_key.parse().expect("Failed to parse private key");
-        let wallet = EthereumWallet::from(signer.clone());
-
         let wallet_address = signer.address();
+        let wallet = EthereumWallet::from(signer);
```
apoorvsadana, line 90: good avoidance of `clone`

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -1,87 +1,135 @@
...
 
         let provider = Arc::new(
             ProviderBuilder::new().with_recommended_fillers().wallet(wallet.clone()).on_http(settlement_cfg.rpc_url),
         );
         let core_contract_client = StarknetValidityContractClient::new(
-            Address::from_str(&settlement_cfg.core_contract_address)
-                .expect("Failed to convert the validity contract address.")
-                .0
-                .into(),
+            Address::from_str(&settlement_cfg.core_contract_address).unwrap().0.into(),
```
apoorvsadana, line 96: why remove `.expect`?

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -1,87 +1,135 @@
...
+        let private_key = get_env_var_or_panic(ENV_PRIVATE_KEY);
+        let signer: PrivateKeySigner = private_key.parse().expect("Failed to parse private key");
+        let wallet_address = signer.address();
+        let wallet = EthereumWallet::from(signer);
+
+        let fill_provider = Arc::new(
+            ProviderBuilder::new().with_recommended_fillers().wallet(wallet.clone()).on_http(settlement_cfg.rpc_url),
+        );
+
+        let core_contract_address = if *SHOULD_IMPERSONATE_ACCOUNT {
```
apoorvsadana, line 116: can you add a small comment explaining the rationale behind this

crates/settlement-clients/ethereum/src/test_data/ABI/starknet_core_contract.json
apoorvsadana, line 1: nitpick: folder name to smallcase `abi`?

crates/settlement-clients/ethereum/src/tests/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+    static ref STARKNET_CORE_CONTRACT_ADDRESS: Address =
+        Address::from_str("0xc662c410c0ecf747543f5ba90660f6abebd9c8c4").expect("Could not impersonate account.");
+}
+
+pub struct TestFixture {
+    pub ethereum_settlement_client: EthereumSettlementClient,
+    pub provider: alloy::providers::RootProvider<alloy::transports::http::Http<reqwest::Client>>,
+}
+
+fn ethereum_test_fixture(block_no: u64) -> TestFixture {
```
apoorvsadana, line 64:
```suggestion
fn setup_ethereum_test(block_no: u64) -> TestSetup {
```

crates/settlement-clients/ethereum/src/tests/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+    let settings_provider: DefaultSettingsProvider = DefaultSettingsProvider {};
+    let ethereum_settlement_client = EthereumSettlementClient::with_test_settings(&settings_provider, provider.clone());
+
+    TestFixture { ethereum_settlement_client, provider }
+}
+
+#[rstest]
+#[tokio::test]
+#[case::basic(20468828)]
+async fn update_state_blob_with_dummy_contract_works(#[case] block_no: u64) {
```
apoorvsadana, line 89: can you also explain the rationale here for having the two test cases

crates/settlement-clients/ethereum/src/tests/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+
+fn get_blob_data(block_no: u64) -> Vec<Vec<u8>> {
+    // Blob Data
+    let blob_data_file_path = format!("{}{}{}{}", *CURRENT_PATH, "/src/test_data/blob_data/", block_no, ".txt");
+    let blob_data = fs::read_to_string(blob_data_file_path).expect("Failed to read the blob data txt file");
+    let blob_data_vec = vec![hex_string_to_u8_vec(&blob_data).unwrap()];
+    blob_data_vec
+}
+
+fn hex_string_to_u8_vec(hex_str: &str) -> color_eyre::Result<Vec<u8>> {
```
apoorvsadana, line 303: it feels like we have a lot of conversion functions now and we might

1. have some duplicates
2. have ways to reduce code using generics (instead of implement a method for both size 32 and size 48, we can use a generic maybe)

not for this PR, maybe we can create an `enhancement` issue

crates/settlement-clients/ethereum/src/tests/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+        .expect("Could not go through update_state_with_blobs.");
+
+    // Asserting, Expected to receive transaction hash.
+    assert!(!update_state_result.is_empty(), "No transaction Hash received.");
+
+    // Call the contract, retrieve the latest stateBlockNumber.
+    let latest_block_number = contract.stateBlockNumber().call().await.unwrap();
+
+    println!("PREVIOUS BLOCK NUMBER {}", prev_block_number._0);
+    println!("CURRENT BLOCK HASH {}", latest_block_number._0);
```
apoorvsadana, line 174:
```suggestion
```

crates/settlement-clients/ethereum/src/tests/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+        .verify_tx_inclusion(update_state_result.as_str())
+        .await
+        .expect("Could not verify inclusion.");
+    assert_eq!(verified_inclusion, SettlementVerificationStatus::Verified);
+}
+
+#[rstest]
+#[tokio::test]
+#[case::typical(20468828, 666039)]
+async fn get_last_settled_block_typical_works(#[case] block_no: u64, #[case] expected: u64) {
```
apoorvsadana, line 193:
```suggestion
async fn get_last_settled_block_typical_works(#[case] fork_block_no: u64, #[case] expected: u64) {
```

apoorvsadana: maybe we can do this for other cases as well

crates/settlement-clients/ethereum/src/tests/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+        if !trimmed.is_empty() {
+            let v_0 = U256::from_str(trimmed).expect("Unable to convert line").to_be_bytes_vec();
+            let v_1 = v_0.as_slice();
+            let v_2 = to_padded_hex(v_1);
+            // let v_3 = v_2.replace("0x", "");
+            let v_4 = hex_string_to_u8_vec(&v_2).expect("unable to convert");
+            let v_5: [u8; 32] = v_4.try_into().expect("Vector length must be 32");
+            program_output.push(v_5)
+        }
+    }
```
apoorvsadana, line 229: we can use the util function below?

crates/settlement-clients/ethereum/src/tests/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+    assert_eq!(input_bytes, expected);
+}
+
+// SOLIDITY FUNCTIONS NEEDED
```
apoorvsadana, line 248: let's move these to the top?

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -537,7 +537,7 @@ pub mod test {
         }
     }
 
-    fn vec_u8_to_hex_string(data: &[u8]) -> String {
+    pub fn vec_u8_to_hex_string(data: &[u8]) -> String {
```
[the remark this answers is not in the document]
heemankv: My apologies, this is not needed, removed.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Added tests for state update job.
 - Tests for DA job.
 - Database tests
+- Tests for Settlement client.
```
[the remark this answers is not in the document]
heemankv: Noted, implemented.

crates/orchestrator/src/jobs/state_update_job/mod.rs
```diff
@@ -62,15 +62,18 @@ impl Job for StateUpdateJob {
             block_numbers = block_numbers.into_iter().filter(|&block| block >= last_failed_block).collect::<Vec<u64>>();
         }
 
+        let mut nonce = config.settlement_client().get_nonce().await?;
+
         let mut sent_tx_hashes: Vec<String> = Vec::with_capacity(block_numbers.len());
         for block_no in block_numbers.iter() {
             let snos = self.fetch_snos_for_block(*block_no).await;
-            let tx_hash = self.update_state_for_block(config, *block_no, snos).await.map_err(|e| {
+            let tx_hash = self.update_state_for_block(config, *block_no, snos, nonce).await.map_err(|e| {
```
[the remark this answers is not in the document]
heemankv: Alright, added.

crates/orchestrator/src/tests/jobs/state_update_job/mod.rs
```diff
@@ -76,14 +76,17 @@ async fn test_process_job_works(
     // functions while fetching the blob data from storage client.
     TestConfigBuilder::new().build().await;
 
+    let nonce: u64 = 3;
```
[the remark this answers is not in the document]
heemankv: `3` here is an arbitrary number,
if we look closely : 
`test_process_job_works` uses `nonce` just to write `expect_method_name` for a mocked settlement client,
which means that nonce ideally is never checked against, hence supplying any `u64` `nonce` works.
Adding a comment about the same.

crates/settlement-clients/ethereum/src/conversion.rs
heemankv: Makes sense, this is cleaner

crates/settlement-clients/ethereum/src/conversion.rs
heemankv: We can, but how do you think the naming of the function be, right now there is a clear decoupling between failing and passing cases naming separating two types of tests.

WDYT ?

crates/settlement-clients/ethereum/src/conversion.rs
heemankv: ### Removing the function : 
This function is a direct dependency for the below two functions, so we cannot remove this function right now/ or merge it some some other function.
Screenshot 2024-08-16 at 11 16 25 AM

### Renaming the function : 
This function converts an array of array of u8s into a Vector of u256, i.e it concats the inner array elements and and forms U256, so we can do as suggested :  `vec_u8_32_to_vec_u256`.

Implemented.

crates/settlement-clients/ethereum/src/conversion.rs
heemankv: We can, but how do you think the naming of the function be, right now there is a clear decoupling between failing and passing cases naming separating two types of tests.

WDYT ?

crates/settlement-clients/ethereum/src/conversion.rs
heemankv: Valid, Added.

crates/settlement-clients/ethereum/src/lib.rs
heemankv: Oh, thanks!

crates/settlement-clients/ethereum/src/lib.rs
heemankv: You are correct, this was removed by mistake, added.

crates/settlement-clients/ethereum/src/lib.rs
heemankv: Added.

crates/settlement-clients/ethereum/src/test_data/ABI/starknet_core_contract.json
heemankv: Corrected.

crates/settlement-clients/ethereum/src/tests/mod.rs
heemankv: Alright, Implemented

crates/settlement-clients/ethereum/src/tests/mod.rs
heemankv: Each function tests different code flow of `update_state_blob` method, 
1) `with_dummy_contract` : tests if the method is able to do a transaction with same function selector on a dummy contract.
2) `with_impersonation` : tests if the method is able to impersonate the`Starknet Operator` and do an `update_state` transaction.

Added as docs for tests

crates/settlement-clients/ethereum/src/tests/mod.rs
heemankv: Noted. 
you can find the issue here:  #84

crates/settlement-clients/ethereum/src/tests/mod.rs
heemankv: Valid, removed.

crates/settlement-clients/ethereum/src/tests/mod.rs
heemankv: Valid, implemented.

crates/settlement-clients/ethereum/src/tests/mod.rs
heemankv: Valid, fixed.

crates/settlement-clients/ethereum/src/tests/mod.rs
heemankv: Valid, moved.

crates/settlement-clients/ethereum/src/conversion.rs
apoorvsadana: hmm, I guess we can do something like `slice_u8_to_u256_all_cases` or `slice_u8_to_u256_working_and_failing_cases`

crates/settlement-clients/ethereum/src/conversion.rs
apoorvsadana: > so we cannot remove this function right now/ or merge it some some other function.

but it's just one line, why can't we write this line over there? for me, it seems it's complicating the code rather than simplifying

crates/settlement-clients/ethereum/src/conversion.rs
apoorvsadana: same as above

crates/settlement-clients/ethereum/src/conversion.rs
apoorvsadana: can we also add a blobscan/etherscan link with actual blob 🙏

crates/settlement-clients/ethereum/src/test_data/ABI/starknet_core_contract.json
apoorvsadana: is it pushed?

crates/settlement-clients/ethereum/src/tests/mod.rs
apoorvsadana: I see, I actually meant the rationale we had initially that when we impersonate we are not checking the signing logic and that's why we wrote this test

crates/settlement-clients/ethereum/src/conversion.rs
heemankv: Alright, Implemented.

crates/settlement-clients/ethereum/src/conversion.rs
heemankv: My experience dictates that it's perfectly normal and recommended to have :
1) Pure functions even if they are used twice, it ensures less area for mistakes by having the code at single place.
2) Clear name dictating what the function is able to perform, the suggested name. `vec_u8_32_to_vec_u256` satisfies that.

Even if you want it to be removed, it's your final call.

crates/settlement-clients/ethereum/src/conversion.rs
heemankv: Alright, Implemented

crates/settlement-clients/ethereum/src/conversion.rs
heemankv: Alright, Added.

crates/settlement-clients/ethereum/src/test_data/ABI/starknet_core_contract.json
heemankv: Yes, Implemented now.

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -1,69 +1,100 @@
...
-use crate::types::EthHttpProvider;
+use crate::conversion::{slice_u8_to_u256, vec_u8_32_to_vec_u256};
+pub mod clients;
+pub mod config;
+pub mod conversion;
+
+// IMPORTANT to understand #[cfg(test)], #[cfg(not(test))] and SHOULD_IMPERSONATE_ACCOUNT
+// Two tests :  `update_state_blob_with_dummy_contract_works` & `update_state_blob_with_impersonation_works` use a env var `TEST_IMPERSONATE_OPERATOR` to inform the function `update_state_with_blobs` about the kind of testing,
+// `TEST_IMPERSONATE_OPERATOR` can have any of "0" or "1" value :
+//      - if "0" then : Testing against Dummy Contract.
```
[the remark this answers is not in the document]
heemankv: Alright, Implemented.

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -1,69 +1,100 @@
...
+//      - if "1" then : Testing via impersonating `Starknet Operator Address`.
+// Note : changing between "0" and "1" is handled automatically by each test function, `no` manual change in `env.test` is needed.
+
+#[cfg(test)]
+lazy_static! {
+    static ref SHOULD_IMPERSONATE_ACCOUNT: bool = get_env_var_or_panic("TEST_IMPERSONATE_OPERATOR") == *"1";
```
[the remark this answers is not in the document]
heemankv: Implemented to be "true" or "false"

crates/settlement-clients/ethereum/src/tests/mod.rs
heemankv: valid, Added.

crates/settlement-clients/ethereum/src/conversion.rs
apoorvsadana: ok let's keep it

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -79,9 +110,36 @@ impl EthereumSettlementClient {
...
+        let private_key = get_env_var_or_panic(ENV_PRIVATE_KEY);
+        let signer: PrivateKeySigner = private_key.parse().expect("Failed to parse private key");
+        let wallet_address = signer.address();
+        let wallet = EthereumWallet::from(signer);
+
+        let fill_provider = Arc::new(
+            ProviderBuilder::new().with_recommended_fillers().wallet(wallet.clone()).on_http(settlement_cfg.rpc_url),
+        );
+
+        let core_contract_address = if *SHOULD_IMPERSONATE_ACCOUNT {
```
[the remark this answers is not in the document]
heemankv: TL:DR : will destroy `single source of truth practice of code`.

`SHOULD_IMPERSONATE_ACCOUNT` is used in `with_test_settings` and `update_state_with_blob`.

If we try this way, we'll be sending a new param to `with_test_settings` which is alright since it's testing code, 
and we don't change anything on `update_state_with_blob` since it's production code, what we end up with is two functions requiring same data but taking it from two difference source, resulting in multiple source of same truth, leading more management for the same.

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -1,69 +1,100 @@
...
     wallet: EthereumWallet,
     wallet_address: Address,
+    #[cfg(not(test))]
+    provider: Arc<EthHttpProvider>,
+    #[cfg(test)]
+    provider: RootProvider<Http<Client>>,
 }
 
 impl EthereumSettlementClient {
+    #[cfg(not(test))]
```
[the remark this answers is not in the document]
heemankv: I think this is depended on : https://github.com/madara-alliance/madara-orchestrator/pull/81#discussion_r1720679752

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -1,69 +1,100 @@
...
+// IMPORTANT to understand #[cfg(test)], #[cfg(not(test))] and SHOULD_IMPERSONATE_ACCOUNT
+// Two tests :  `update_state_blob_with_dummy_contract_works` & `update_state_blob_with_impersonation_works` use a env var `SHOULD_IMPERSONATE_ACCOUNT` to inform the function `update_state_with_blobs` about the kind of testing,
+// `SHOULD_IMPERSONATE_ACCOUNT` can have any of "0" or "1" value :
+//      - if "0" then : Testing via default Anvil address.
+//      - if "1" then : Testing via impersonating `Starknet Operator Address`.
+// Note : changing between "0" and "1" is handled automatically by each test function, `no` manual change in `env.test` is needed.
+
+#[cfg(test)]
+lazy_static! {
+    static ref SHOULD_IMPERSONATE_ACCOUNT: bool = get_env_var_or_panic("SHOULD_IMPERSONATE_ACCOUNT") == *"true";
```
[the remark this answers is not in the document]
heemankv: Valid thought process, 
Have a look at the cleaner implementation here : [Commit](https://github.com/madara-alliance/madara-orchestrator/pull/81/commits/a8f4bb16a3fc5b2c911c21c3a03e8bdbace8eaf7)

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -174,20 +247,30 @@ impl SettlementClient for EthereumSettlementClient {
...
-        let encoded = tx_envelope.encoded_2718();
+        // IMP: this conversion strips signature from the transaction
 
-        let pending_tx = self.provider.send_raw_transaction(&encoded).await?;
+        #[cfg(not(test))]
+        let txn_request: TransactionRequest = tx_envelope.into();
 
-        Ok(pending_tx.tx_hash().to_string())
+        #[cfg(test)]
+        let mut txn_request: TransactionRequest = tx_envelope.into();
```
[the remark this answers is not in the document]
heemankv: Valid thought process, 
Have a look at the cleaner implementation here : [Commit](https://github.com/madara-alliance/madara-orchestrator/pull/81/commits/a8f4bb16a3fc5b2c911c21c3a03e8bdbace8eaf7)

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -79,9 +110,36 @@ impl EthereumSettlementClient {
...
+        let wallet = EthereumWallet::from(signer);
+
+        let fill_provider = Arc::new(
+            ProviderBuilder::new().with_recommended_fillers().wallet(wallet.clone()).on_http(settlement_cfg.rpc_url),
+        );
+
+        let core_contract_address = if *SHOULD_IMPERSONATE_ACCOUNT {
+            &settlement_cfg.core_contract_address
+        } else {
+            &*TEST_DUMMY_CONTRACT_ADDRESS
```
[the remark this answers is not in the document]
heemankv: Answered [here](https://github.com/madara-alliance/madara-orchestrator/pull/81#discussion_r1720920880)

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -1,69 +1,100 @@
...
 #[allow(dead_code)]
 pub struct EthereumSettlementClient {
-    provider: Arc<EthHttpProvider>,
     core_contract_client: StarknetValidityContractClient,
     wallet: EthereumWallet,
     wallet_address: Address,
+    #[cfg(not(test))]
+    provider: Arc<EthHttpProvider>,
+    #[cfg(test)]
+    provider: RootProvider<Http<Client>>,
```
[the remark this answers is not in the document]
heemankv: Valid, I have removed the usage of `EthHttpProvider` now,
here's the [commit](https://github.com/madara-alliance/madara-orchestrator/pull/81/commits/30e85844516d23438c236e35a940718777fe7d3f)

crates/settlement-clients/ethereum/src/lib.rs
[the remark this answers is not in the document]
heemankv: Update: removed.

crates/settlement-clients/ethereum/src/lib.rs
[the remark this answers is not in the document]
apoorvsadana: looks good

apoorvsadana (approved):

crates/settlement-clients/ethereum/src/conversion.rs
```diff
@@ -1,12 +1,304 @@
+use alloy::dyn_abi::parser::Error;
+use alloy::eips::eip4844::BYTES_PER_BLOB;
 use alloy::primitives::U256;
+use alloy_primitives::FixedBytes;
+use c_kzg::{Blob, KzgCommitment, KzgProof, KzgSettings};
+use color_eyre::{eyre::ContextCompat, Result as EyreResult};
+use std::fmt::Write;
 
 /// Converts a `&[Vec<u8>]` to `Vec<U256>`. Each inner slice is expected to be exactly 32 bytes long.
```
odesenfans, line 9: This comment is slightly wrong as you're taking `&[[u8; 32]]`. Update it?

crates/settlement-clients/ethereum/src/conversion.rs
```diff
@@ -1,12 +1,304 @@
...
 /// Converts a `&[u8]` to `U256`.
-pub(crate) fn slice_u8_to_u256(slice: &[u8]) -> U256 {
-    U256::try_from_be_slice(slice).expect("could not convert u8 slice to U256")
+pub(crate) fn slice_u8_to_u256(slice: &[u8]) -> EyreResult<U256> {
+    U256::try_from_be_slice(slice).wrap_err_with(|| "could not convert &[u8] to U256".to_string())
+}
+
+// Function to convert a slice of u8 to a padded hex string
+// Function only takes a slice of length up to 32 elements
+// Pads the value on the right side with zeros only if the converted string has lesser than 64 characters.
```
odesenfans, line 22: Use docstrings to document functions (`///` instead of `//`).

crates/settlement-clients/ethereum/src/conversion.rs
```diff
@@ -1,12 +1,304 @@
...
+        + &length_program_output_hex
+        + &lines_count_hex
+        + &program_output_hex
+        + &length_kzg_hex
+        + &kzg_proof_hex;
+
+    Ok(input_data)
+}
+
+pub(crate) fn vec_u8_32_to_hex_string(data: Vec<[u8; 32]>) -> String {
```
odesenfans, line 75: Why do you need ownership here? Can't you pass a slice in all cases?

crates/settlement-clients/ethereum/src/tests/mod.rs
```diff
@@ -0,0 +1,341 @@
...
+    }
+}
+
+struct EthereumTestBuilder {
+    fork_block: Option<u64>,
+    impersonator: Option<Address>,
+}
+
+struct EthereumTest {
+    _anvil: AnvilInstance,
```
odesenfans, line 82: Why are you prefixing this with an underscore?

odesenfans (commented):
I would need to understand the context a bit more to review this PR meaningfully. I added a few comments but they're not super interesting.

.env.test
```diff
@@ -15,7 +15,7 @@ MADARA_RPC_URL="http://localhost:3000"
 ETHEREUM_RPC_URL="http://localhost:3001"
 MEMORY_PAGES_CONTRACT_ADDRESS="0x000000000000000000000000000000000001dead"
 PRIVATE_KEY="0xdead"
-ETHEREUM_PRIVATE_KEY="0x000000000000000000000000000000000000000000000000000000000000beef"
+ETHEREUM_PRIVATE_KEY="0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"
```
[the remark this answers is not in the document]
heemankv: Ideally, we can remove this but it's genuinely no harm since this is Private_key provided by Anvil, will add a comment to explain the same

crates/settlement-clients/ethereum/Cargo.toml
```diff
@@ -4,11 +4,14 @@ version.workspace = true
 edition.workspace = true
 
 [dependencies]
-alloy = { workspace = true, features = ["full"] }
+alloy = { workspace = true, features = ["full", "node-bindings"] }
+alloy-primitives = { version = "0.7.7", default-features = false }
```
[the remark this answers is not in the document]
heemankv: Created an issue for the same to be thoroughly checked later on :  #99

crates/settlement-clients/ethereum/src/conversion.rs
heemankv: Valid, updated.

crates/settlement-clients/ethereum/src/conversion.rs
heemankv: Valid, improved. using ` &[[u8; 32]]` now

crates/settlement-clients/ethereum/src/tests/mod.rs
heemankv: So, 
We notices the an `anvil` server is needed to be in scope in any function that is using RPC calls to `anvil`.

We create `anvil` server in `EthereumTestBuilder` if we don't pass it to `update_state_blob_with_dummy_contract_works` we get `ConnectionRefused` Error suggesting that the server that was started in `EthereumTestBuilder` has now been closed.

I believe this has to do with how `rust` handles `variable scoping`.

So, `_anvil` as a variable is never used, it's only there to be in the scope, hence the `_`.

This is my understanding, please feel free to correct it / broaden the perspective / suggest an alternative & optimised way to achieve the same.

PR #83, opened 2024-08-13
Chore : Renaming Tests & No test returns

heemankv:
# This PR ensures that tests follow correct test conventions.

### No test function should have a return statement, there's no benefit in getting a test to return any value.


apoorvsadana:
Closing this for now, will pick this later.

Issue #68, opened 2024-08-01
remove `init_config` from tests and use `TestConfigBuilder`


apoorvsadana:
We don't need to use `init_config` anymore. `TestConfigBuilder` should be able to replace it now.

elielnfinic:
Can I work on this @apoorvsadana ?

apoorvsadana:
Hey @elielnfinic, thanks for reaching out. Currently, there are four teams working on picking issues and we're trying to keep it to that for now. The main reason being to be more accountable and to be able to ship fast. So for now, we haven't opened issues to new contributors.

elielnfinic:
Sounds good. Thank you.

ocdbytes:
Ig we can close this @apoorvsadana

PR #87, opened 2024-08-17
Feat : removed init_config function from tests

ocdbytes:
- Removed `init_config` function from tests.
- Issue : #68


apoorvsadana (approved):

antiyro (approved):
Seems good to me, the PR is well written and reduce a lot of overheads and default configs while keeping it clear.

Issue #86, opened 2024-08-17
Feat : Updating the logic of fetching of PIE from zip file in local to remote storage


ocdbytes:
- Current scenario :
    Right now we are reading the PIE file from the local directory during the execution of the proving job.
    ```rs
    async fn process_job(&self, config: &Config, job: &mut JobItem) -> Result<String, JobError> {
        // TODO: allow to download PIE from storage
        let cairo_pie_path = job
            .metadata
            .get(JOB_METADATA_CAIRO_PIE_PATH_KEY)
            .map(|s| PathBuf::from_str(s))
            .ok_or_else(|| ProvingError::CairoPIEWrongPath { internal_id: job.internal_id.clone() })?
            .map_err(|_| ProvingError::CairoPIENotReadable)?;

        let cairo_pie = CairoPie::read_zip_file(&cairo_pie_path).map_err(|_| ProvingError::CairoPIENotReadable)?;

        let external_id = config
            .prover_client()
            .submit_task(Task::CairoPie(cairo_pie))
            .await
            .wrap_err("Prover Client Error".to_string())
            .map_err(|e| JobError::Other(OtherError(e)))?;

        Ok(external_id)
    }
    ```

- Ideal Scenario : 
    We need to implement the logic in which we will read the PIE file from remote storage such as AWS S3 and then send the cairo pie from there.

ocdbytes:
Issue resolved in PR : #88

PR #88, opened 2024-08-17
Feat: proving job storage integration

ocdbytes:
- Added PIE file fetch from storage client.
- Issue #86

### DIFF
--- a/crates/orchestrator/src/jobs/proving_job/mod.rs
+++ b/crates/orchestrator/src/jobs/proving_job/mod.rs
@@ -1,6 +1,4 @@
 use std::collections::HashMap;
-use std::path::PathBuf;
-use std::str::FromStr;
 
 use async_trait::async_trait;
 use cairo_vm::vm::runners::cairo_pie::CairoPie;
@@ -11,7 +9,6 @@ use tracing::log::log;
 use tracing::log::Level::Error;
 use uuid::Uuid;
 
-use super::constants::JOB_METADATA_CAIRO_PIE_PATH_KEY;
 use super::types::{JobItem, JobStatus, JobType, JobVerificationStatus};
 use super::{Job, JobError, OtherError};
 use crate::config::Config;
@@ -38,10 +35,6 @@ impl Job for ProvingJob {
         internal_id: String,
         metadata: HashMap<String, String>,
     ) -> Result<JobItem, JobError> {
-        if !metadata.contains_key(JOB_METADATA_CAIRO_PIE_PATH_KEY) {
-            // TODO: validate the usage of `.clone()` here, ensure lightweight borrowing of variables
-            Err(ProvingError::CairoPIEWrongPath { internal_id: internal_id.clone() })?
-        }
         Ok(JobItem {
             id: Uuid::new_v4(),
             internal_id,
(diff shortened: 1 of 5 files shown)


crates/orchestrator/src/jobs/proving_job/mod.rs
```diff
@@ -54,15 +47,12 @@ impl Job for ProvingJob {
...
-            .get(JOB_METADATA_CAIRO_PIE_PATH_KEY)
-            .map(|s| PathBuf::from_str(s))
-            .ok_or_else(|| ProvingError::CairoPIEWrongPath { internal_id: job.internal_id.clone() })?
-            .map_err(|_| ProvingError::CairoPIENotReadable)?;
-
-        let cairo_pie = CairoPie::read_zip_file(&cairo_pie_path).map_err(|_| ProvingError::CairoPIENotReadable)?;
+        // Cairo Pie path in s3 storage client
+        let cairo_pie_path = job.internal_id.to_string() + "/pie.zip";
+        let cairo_pie_file =
+            config.storage().get_data(&cairo_pie_path).await.map_err(|_| ProvingError::CairoPIENotReadable)?;
```
apoorvsadana, line 53: let's pass the error object into the function

crates/orchestrator/src/jobs/proving_job/mod.rs
```diff
@@ -38,10 +38,6 @@ impl Job for ProvingJob {
         internal_id: String,
         metadata: HashMap<String, String>,
     ) -> Result<JobItem, JobError> {
-        if !metadata.contains_key(JOB_METADATA_CAIRO_PIE_PATH_KEY) {
```
apoorvsadana, line 41: we can delete this from constants folder as well

crates/orchestrator/src/jobs/proving_job/mod.rs
```diff
@@ -54,15 +50,15 @@ impl Job for ProvingJob {
...
-        let cairo_pie = CairoPie::read_zip_file(&cairo_pie_path).map_err(|_| ProvingError::CairoPIENotReadable)?;
+        // Cairo Pie path in s3 storage client
+        let cairo_pie_path = job.internal_id.to_string() + "/pie.zip";
+        let cairo_pie_file = config
+            .storage()
+            .get_data(&cairo_pie_path)
+            .await
+            .map_err(|e| ProvingError::CairoPIEFileFetchFailed(e.to_string()))?;
+        let cairo_pie =
+            CairoPie::from_bytes(cairo_pie_file.to_vec().as_slice()).map_err(|_| ProvingError::CairoPIENotReadable)?;
```
apoorvsadana, line 61: still not reading `e` over here? what's the confusion?

ocdbytes: we talked about this right ? sirf upar wale me e read krna tha. neeche waale me you told that we need to pass CairoPieNoReadable

apoorvsadana: No. It was that we can do `Other` error  (we're using `CairoPIEFileFetchFailed` now but that's fine) on the first one and `CairoPieNotReadable` in the lower. As a thumb rule, we should always pass the error objects because if this breaks tomorrow, we won't have context of the actual error.

apoorvsadana (approved):

EvolveArt (approved):

PR #89, opened 2024-08-18
small changes

apoorvsadana:


crates/settlement-clients/ethereum/src/tests/mod.rs
```diff
@@ -77,61 +73,83 @@ sol! {
...
-    assert_eq!(
-        contract.address().to_string(),
-        *TEST_DUMMY_CONTRACT_ADDRESS,
-        "Dummy Contract got deployed on unexpected address"
+    let contract = DummyCoreContract::deploy(&setup.provider).await.expect("Unable to deploy address");
+    let ethereum_settlement_client = EthereumSettlementClient::with_test_settings(
+        setup.provider.clone(),
+        Some(*contract.address()),
+        setup.rpc_url,
+        None,
```
heemankv, line 143: Why are we not sending the `DummyContractAddress` here ? and sending `None` instead.

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -87,46 +73,63 @@ impl EthereumSettlementClient {
...
+            Arc::new(ProviderBuilder::new().with_recommended_fillers().wallet(wallet.clone()).on_http(rpc_url));
 
-        let core_contract_address = if *SHOULD_IMPERSONATE_ACCOUNT {
-            &settlement_cfg.core_contract_address
+        let core_contract_address = if let Some(core_contract_address) = core_contract_address {
+            core_contract_address.clone()
         } else {
-            &*TEST_DUMMY_CONTRACT_ADDRESS
+            // dummy address
+            Address::from_str("0x0").unwrap()
```
heemankv, line 121: If the dummy contract Address is `0x0` how will the dummy transaction be made ?
Moreover Why are we not sending the DummyContractAddress in mod.rs line 143 ? and sending None instead.

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -297,14 +295,25 @@ impl SettlementClient for EthereumSettlementClient {
...
-            txn_request = txn_request.with_from(*ADDRESS_TO_IMPERSONATE);
+        // IMPORTANT to understand #[cfg(test)], #[cfg(not(test))] and SHOULD_IMPERSONATE_ACCOUNT
+        // Two tests :  `update_state_blob_with_dummy_contract_works` & `update_state_blob_with_impersonation_works` use a env var `SHOULD_IMPERSONATE_ACCOUNT` to inform the function `update_state_with_blobs` about the kind of testing,
+        // `SHOULD_IMPERSONATE_ACCOUNT` can have any of "0" or "1" value :
+        //      - if "0" then : Testing via default Anvil address.
+        //      - if "1" then : Testing via impersonating `Starknet Operator Address`.
+        // Note : changing between "0" and "1" is handled automatically by each test function, `no` manual change in `env.test` is needed.
+        if let Some(impersonate_account) = impersonate_account {
+            let nonce =
+                provider.get_transaction_count(impersonate_account).await.unwrap().to_string().parse::<u64>().unwrap();
```
heemankv, line 314: We already send the `nonce` to the `update_state_with_blobs` method then why are we recalculating it here ?

.env.test
```diff
@@ -28,7 +28,6 @@ MONGODB_CONNECTION_STRING="mongodb://localhost:27017"
 
 
 # Ethereum Settlement
-DEFAULT_SETTLEMENT_CLIENT_RPC="http://localhost:3000"
```
heemankv, line 31: It is valid to remove this, since the `rpc_url` is being sent by anvil as a param to the child methods

PR #92, opened 2024-08-20
feat : added worker queues and listeners to consume the message

ocdbytes:
- Added queues for worker trigger listeners.
- Added consumer spawn functions for worker listeners.


crates/orchestrator/src/queue/job_queue.rs
```diff
@@ -52,60 +83,134 @@ where
...
     log::info!("Consuming from queue {:?}", queue);
-    let config = config().await;
-    let delivery = match config.queue().consume_message_from_queue(queue.clone()).await {
-        Ok(d) => d,
-        Err(QueueError::NoData) => {
-            return Ok(());
-        }
-        Err(e) => {
-            return Err(ConsumptionError::FailedToConsumeFromQueue { error_msg: e.to_string() });
+    let delivery = get_delivery_from_queue(queue.clone()).await?;
```
apoorvsadana, line 86: let's use reference here

crates/orchestrator/src/queue/job_queue.rs
```diff
@@ -52,60 +83,134 @@ where
...
+where
+    F: FnOnce(Box<dyn Worker>) -> Fut,
+    Fut: Future<Output = color_eyre::Result<()>>,
+{
+    log::info!("Consuming from queue {:?}", queue);
+    let delivery = get_delivery_from_queue(queue.clone()).await?;
+
+    match delivery {
+        DeliveryReturnType::Message(d) => {
+            let job_message: Option<WorkerTriggerMessage> = d
```
apoorvsadana, line 147:
```suggestion
            let event_trigger: Option<WorkerTriggerMessage> = d
```

crates/orchestrator/src/queue/job_queue.rs
```diff
@@ -198,8 +198,8 @@ fn get_worker_handler_from_worker_trigger_type(worker_trigger_type: WorkerTrigge
 }
 
 /// To get the delivery from the message queue using the queue name
-async fn get_delivery_from_queue(queue: String) -> Result<DeliveryReturnType, ConsumptionError> {
-    match config().await.queue().consume_message_from_queue(queue.clone()).await {
+async fn get_delivery_from_queue(queue: &str) -> Result<DeliveryReturnType, ConsumptionError> {
+    match config().await.queue().consume_message_from_queue(queue.to_string()).await {
```
apoorvsadana, line 202: the `.to_string()` actually beats the purpose because it's still being cloned. can we actually optimise the `consume_message_from_queue` method to work on &str or a reference instead?

apoorvsadana (approved):

crates/orchestrator/src/queue/job_queue.rs
```diff
@@ -52,60 +83,134 @@ where
...
+    }
+
+    Ok(())
+}
+
+/// Function to consume the message from the worker trigger queues and spawn the worker
+/// for respective message received.
+pub async fn consume_worker_trigger_messages_from_queue<F, Fut>(
+    queue: String,
+    handler: F,
```
odesenfans, line 136: A design question I have: why is the handler async if you're going to `await` it anyway? It seems like you're not taking advantage of the asynchronicity here by running multiple workers in parallel. Is it something you'll support later? And/or do you need the handler to be async because it calls async functions itself?

crates/orchestrator/src/queue/sqs/mod.rs
```diff
@@ -1,15 +1,30 @@
...
 use crate::queue::QueueProvider;
 pub struct SqsQueue;
 
+lazy_static! {
+    /// Maps Queue Name to Env var of queue URL.
+    pub static ref QUEUE_NAME_TO_ENV_VAR_MAPPING: HashMap<String, String> = HashMap::from([
+        (JOB_PROCESSING_QUEUE.to_string(), "SQS_JOB_PROCESSING_QUEUE_URL".to_string()),
+        (JOB_VERIFICATION_QUEUE.to_string(), "SQS_JOB_VERIFICATION_QUEUE_URL".to_string()),
+        (JOB_HANDLE_FAILURE_QUEUE.to_string(), "SQS_JOB_HANDLE_FAILURE_QUEUE_URL".to_string()),
+        (WORKER_TRIGGER_QUEUE.to_string(), "SQS_WORKER_TRIGGER_QUEUE_URL".to_string()),
```
odesenfans, line 24: I don't see any reason to convert these to strings, seems like you could be happy with a `HashMap<&str, &str>`?

odesenfans (changes requested):
Some code styling and a question on the design.

crates/orchestrator/src/queue/job_queue.rs
ocdbytes: Here technically all the queues are consuming the messages in parellel only or I didn't understand your question. 

Here I am adding the consumers : 

```rust
macro_rules! spawn_consumer {
    ($queue_type :expr, $handler : expr, $consume_function: expr) => {
        tokio::spawn(async move {
            loop {
                match $consume_function($queue_type, $handler).await {
                    Ok(_) => {}
                    Err(e) => log::error!("Failed to consume from queue {:?}. Error: {:?}", $queue_type, e),
                }
            }
        });
    };
}

pub async fn init_consumers() -> Result<(), JobError> {
    spawn_consumer!(JOB_PROCESSING_QUEUE.to_string(), process_job, consume_job_from_queue);
    spawn_consumer!(JOB_VERIFICATION_QUEUE.to_string(), verify_job, consume_job_from_queue);
    spawn_consumer!(JOB_HANDLE_FAILURE_QUEUE.to_string(), handle_job_failure, consume_job_from_queue);
    spawn_consumer!(WORKER_TRIGGER_QUEUE.to_string(), spawn_worker, consume_worker_trigger_messages_from_queue);
    Ok(())
}
```

Can you explain more on this ? that would be helpful. 
Thanks

crates/orchestrator/src/queue/job_queue.rs
```diff
@@ -52,60 +83,134 @@ where
...
+                                    worker_trigger_type: job_message.worker,
+                                    error_msg: "Worker handling failed, message nack-ed".to_string(),
+                                })?,
+                                Err(delivery_nack_error) => Err(ConsumptionError::FailedToSpawnWorker {
+                                    worker_trigger_type: job_message.worker,
+                                    error_msg: delivery_nack_error.0.to_string(),
+                                })?,
+                            }
+                        }
+                    };
```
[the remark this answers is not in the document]
ocdbytes: done

crates/orchestrator/src/queue/job_queue.rs
```diff
@@ -52,60 +83,134 @@ where
...
+) -> Result<(), ConsumptionError>
+where
+    F: FnOnce(Box<dyn Worker>) -> Fut,
+    Fut: Future<Output = color_eyre::Result<()>>,
+{
+    log::info!("Consuming from queue {:?}", queue);
+    let delivery = get_delivery_from_queue(&queue).await?;
+
+    match delivery {
+        DeliveryReturnType::Message(d) => {
```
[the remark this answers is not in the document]
ocdbytes: renamed

crates/orchestrator/src/queue/sqs/mod.rs
ocdbytes: done

ocdbytes:
Requesting review @odesenfans 🙏

crates/orchestrator/src/queue/job_queue.rs
odesenfans: Yes, I saw that you spawned multiple consumers, one per queue. My remark is that each queue consumer is sequential, i.e. all messages coming from one queue are treated one after the other here. There's no inherent problem with that, I was just wondering if this was the design you were going for or if you wanted to treat messages from each queue in parallel in the future as well.

odesenfans (approved):

crates/orchestrator/src/queue/job_queue.rs
ocdbytes: Good point I will create an issue for this and we will process all the queue messages in parallel

PR #101, opened 2024-08-27
Tests : e2e test

ocdbytes:


crates/orchestrator/src/data_storage/aws_s3/mod.rs
```diff
@@ -60,7 +60,6 @@ impl DataStorage for AWSS3 {
         Ok(())
     }
 
-    #[cfg(test)]
```
apoorvsadana, line 63: why do we need to remove this?

crates/orchestrator/src/data_storage/mod.rs
```diff
@@ -17,7 +17,6 @@ use mockall::automock;
 pub trait DataStorage: Send + Sync {
     async fn get_data(&self, key: &str) -> Result<Bytes>;
     async fn put_data(&self, data: Bytes, key: &str) -> Result<()>;
-    #[cfg(test)]
```
apoorvsadana, line 20: same as above

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -198,30 +203,28 @@ pub fn convert_to_biguint(elements: Vec<FieldElement>) -> Vec<BigUint> {
     biguint_vec
 }
 
-fn data_to_blobs(blob_size: u64, block_data: Vec<BigUint>) -> Result<Vec<Vec<u8>>, JobError> {
+pub fn data_to_blobs(blob_size: u64, block_data: Vec<BigUint>) -> Result<Vec<Vec<u8>>, JobError> {
```
apoorvsadana, line 206: can we remove `pub`?

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -198,30 +203,28 @@ pub fn convert_to_biguint(elements: Vec<FieldElement>) -> Vec<BigUint> {
...
     if blob_size < 32 {
         Err(DaError::InsufficientBlobSize { blob_size })?
     }
 
     let mut blobs: Vec<Vec<u8>> = Vec::new();
 
-    // Convert all FieldElements to bytes first
-    let mut bytes: Vec<u8> = block_data.iter().flat_map(|element| element.to_bytes_be().to_vec()).collect();
+    // Convert all BigUint to bytes more efficiently
+    let bytes: Vec<u8> = block_data.iter().flat_map(BigUint::to_bytes_be).collect();
```
apoorvsadana, line 215:
```suggestion
    let bytes: Vec<u8> = block_data.into_iter().flat_map(BigUint::to_bytes_be).collect();
```

whenever you're iterating over something that you won't use later, use `into_iter` as it moved the value instead of cloning

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -198,30 +203,28 @@ pub fn convert_to_biguint(elements: Vec<FieldElement>) -> Vec<BigUint> {
...
     // Validate blob size
     if blob_size < 32 {
         Err(DaError::InsufficientBlobSize { blob_size })?
     }
 
     let mut blobs: Vec<Vec<u8>> = Vec::new();
 
-    // Convert all FieldElements to bytes first
-    let mut bytes: Vec<u8> = block_data.iter().flat_map(|element| element.to_bytes_be().to_vec()).collect();
+    // Convert all BigUint to bytes more efficiently
```
apoorvsadana, line 214:
```suggestion
    // Convert all BigUint to bytes
```

the more is probably referencing to the replaced code, others won't know that

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -198,30 +203,28 @@ pub fn convert_to_biguint(elements: Vec<FieldElement>) -> Vec<BigUint> {
...
-        blobs.push(last_blob);
-        log::debug!("Warning: Remaining {} bytes not forming a complete blob were padded", remaining_bytes);
+    let chunk_size = blob_size as usize;
+    let chunks = bytes.chunks(chunk_size);
+
+    for chunk in chunks {
+        let mut blob = chunk.to_vec();
+        if blob.len() < chunk_size {
+            blob.resize(chunk_size, 0);
+            log::debug!("Warning: Last chunk of {} bytes was padded to full blob size", chunk.len());
```
apoorvsadana, line 225: should we throw an error if it's not the last?

crates/orchestrator/src/queue/job_queue.rs
```diff
@@ -57,11 +58,42 @@ pub enum WorkerTriggerType {
...
+        }
+    }
+}
+
+impl<'de> Deserialize<'de> for WorkerTriggerMessage {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let s = String::deserialize(deserializer)?;
```
apoorvsadana, line 86: we should be able to use `serde_json`

e2e-tests/src/localstack.rs
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    /// Generic function to send message to any of the queues
+    pub async fn send_message_to_queue(&self, queue_url: &str, message_body: &str) -> color_eyre::Result<()> {
+        let sqs_client = self.sqs_client().await;
+        sqs_client.send_message().queue_url(queue_url).message_body(message_body).send().await?;
+        Ok(())
+    }
+
+    async fn sqs_client(&self) -> aws_sdk_sqs::Client {
```
apoorvsadana, line 169: the clients should be a part of the `LocalStack` struct + the new method which builds the client and returns the struct

e2e-tests/src/localstack.rs
```diff
@@ -0,0 +1,219 @@
...
+        let region_provider = Region::new("us-east-1");
```
apoorvsadana, line 170: should be taken from env var?

e2e-tests/src/mock_server.rs
```diff
@@ -0,0 +1,47 @@
...
+    pub fn url(&self) -> String {
+        self.client_url.clone()
+    }
+
+    /// To add mock on the mock server endpoints
+    pub fn add_mock_on_endpoint(
+        &mut self,
+        path: &str,
+        body_contains: Option<&str>,
+        body_contains_2: Option<&str>,
```
apoorvsadana, line 32: doesn't look like the correct design, what are we trying to do here

e2e-tests/src/localstack.rs
```diff
@@ -0,0 +1,219 @@
...
+            "https://madara-orchestrator-sharp-pie.s3.amazonaws.com/{}-SN.zip",
+            self.l2_block_number
+        ))
+        .await?;
+        let file_bytes = file.bytes().await?;
+
+        // putting the pie file into localstack s3
+        let s3_file_key = self.l2_block_number.to_string() + "/pie.zip";
+        s3_client.put_data(file_bytes, &s3_file_key).await?;
+        println!("✔️ PIE file uploaded to localstack s3");
```
apoorvsadana, line 105: this part is specific to the e2e test case. imaging that the localstack, mongo etc. files are general for all e2e test cases (if we had more)

e2e-tests/src/mongodb.rs
```diff
@@ -1,34 +1,28 @@
...
-        let host_port = get_free_port();
-
-        let container = GenericImage::new(MONGODB_IMAGE_NAME, MONGODB_IMAGE_TAG)
-            .with_wait_for(WaitFor::message_on_stdout("Waiting for connections"))
-            .with_mapped_port(host_port, ContainerPort::Tcp(MONGODB_DEFAULT_PORT))
-            .start()
-            .await
-            .expect("Failed to create docker container");
-        Self { container, endpoint: Url::parse(&format!("http://127.0.0.1:{}", host_port)).unwrap() }
+        // TODO : check why test-containers is not working
```
apoorvsadana, line 13: if we are planning to not use testcontainers then let's remove this

e2e-tests/src/mongodb.rs
```diff
@@ -1,34 +1,28 @@
...
+        //     .start()
+        //     .await
+        //     .expect("Failed to create docker container");
+        Self { endpoint: Url::from_str("mongodb://localhost:27017").unwrap() }
     }
 
-    pub fn endpoint(&self) -> &Url {
-        &self.endpoint
+    pub fn endpoint(&self) -> Url {
+        Url::from_str("mongodb://localhost:27017").unwrap()
```
apoorvsadana, line 26: should be taken from env var

e2e-tests/src/sharp.rs
apoorvsadana, line 1: do we need a separate folder for this? for us it's just the`mock_server`, there's no difference?

e2e-tests/src/starknet_client.rs
apoorvsadana, line 1: same as above

e2e-tests/tests.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    // Run orchestrator
+    let mut orchestrator = Orchestrator::run(env_vec);
+    orchestrator.wait_till_started().await;
+
+    // TODO : need to make this dynamic
+    sleep(Duration::from_secs(900)).await;
+
+    // Adding a case here to check for required state of the orchestrator to end the test.
+    let l2_block_for_testing = get_env_var_or_panic("L2_BLOCK_NUMBER_FOR_TEST");
```
apoorvsadana, line 63: why does this need to be an env var?

.env.test
```diff
@@ -50,8 +51,9 @@ SHARP_PROOF_LAYOUT="small"
 
 DA_LAYER="ethereum"
 SETTLEMENT_LAYER="ethereum"
-SETTLEMENT_RPC_URL="http://localhost:3001"
-MADARA_RPC_URL="http://localhost:3000"
+SETTLEMENT_RPC_URL="https://eth-mainnet.public.blastapi.io"
+MADARA_RPC_URL="http://localhost:5000"
+ETHEREUM_RPC_URL="https://eth-mainnet.public.blastapi.io"
```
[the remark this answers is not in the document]
ocdbytes: it makes sense to remove the `SETTLEMENT_RPC_URL` variable ig because `ETHEREUM_RPC_URL ` is being used in prover client and other places also.

e2e-tests/src/ethereum.rs
```diff
@@ -0,0 +1,65 @@
...
+        println!("♢ Impersonate Account setup completed.");
+    }
+
+    /// To get the anvil endpoint
+    pub fn endpoint(&self) -> String {
+        self.anvil_endpoint.clone()
+    }
+
+    /// To get the signer
+    pub fn get_signer(&self) -> EthereumWallet {
```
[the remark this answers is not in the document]
ocdbytes: this will return the default signer also added  an additional comment here explaining the purpose of this.

crates/orchestrator/src/data_storage/aws_s3/mod.rs
ocdbytes: So basically this function is needed in the e2e tests and tests environment was not available in e2e tests as it is a different crate ig so that's why made it public.

crates/orchestrator/src/data_storage/mod.rs
ocdbytes: https://github.com/madara-alliance/madara-orchestrator/pull/101#discussion_r1741645986

crates/orchestrator/src/jobs/da_job/mod.rs
ocdbytes: got it.

crates/orchestrator/src/jobs/da_job/mod.rs
ocdbytes: Idts we should throw an error in this case, as it is not an error as blob may have length less than the blob size.

crates/orchestrator/src/queue/job_queue.rs
ocdbytes: some error was coming when I was using serde_json that's why I implemented this custom deserializer.

crates/orchestrator/src/workers/update_state.rs
```diff
@@ -24,28 +24,58 @@ impl Worker for UpdateStateWorker {
...
                         latest_successful_job_internal_id,
                     )
                     .await?;
 
+                let successful_da_jobs_without_successor = config
+                    .database()
+                    .get_jobs_without_successor(JobType::DataSubmission, JobStatus::Completed, JobType::StateTransition)
+                    .await?;
+
+                let result_vector = intersection(&successful_da_jobs, &successful_da_jobs_without_successor);
```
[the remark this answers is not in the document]
ocdbytes: Sorry,. My bad I forgot that even If the job comes again the create job function will detect it and will prevent it from being created again.

crates/prover-services/sharp-service/src/client.rs
```diff
@@ -97,10 +94,17 @@ impl SharpClient {
         // Adding params to the url
         add_params_to_url(&mut base_url, params);
 
-        let res = self.client.post(base_url).send().await.map_err(SharpError::GetJobStatusFailure)?;
+        let res = self.client.post(base_url.clone()).send().await.map_err(SharpError::GetJobStatusFailure)?;
 
         match res.status() {
-            reqwest::StatusCode::OK => res.json().await.map_err(SharpError::GetJobStatusFailure),
+            reqwest::StatusCode::OK => {
```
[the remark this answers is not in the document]
ocdbytes: No so the thing is something is breaking It is not able to get the response from the json idk why I spent some time to debug it then gave up on this.

crates/prover-services/sharp-service/src/types.rs
```diff
@@ -14,6 +14,7 @@ pub struct SharpGetProofResponse {
 
 #[derive(Default, Debug, Clone, Deserialize)]
 pub struct SharpGetStatusResponse {
+    #[serde(default)]
```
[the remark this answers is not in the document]
ocdbytes: It is defined like this in snos crate and I am importing the type from there.

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -89,17 +115,9 @@ impl EthereumSettlementClient {
             filler_provider,
         );
 
-        EthereumSettlementClient {
-            provider,
-            core_contract_client,
-            wallet,
-            wallet_address,
-            #[cfg(test)]
-            impersonate_account: None,
```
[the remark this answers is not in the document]
ocdbytes: should we make another issue for this ?

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -237,18 +246,12 @@ impl SettlementClient for EthereumSettlementClient {
...
-
-        #[cfg(test)]
-        let txn_request = test_config::configure_transaction(
-            // self.provider.clone(),
-            tx_envelope,
-            self.impersonate_account,
-            nonce,
-        )
-        .await;
+        let txn_request: TransactionRequest = match self.impersonate_account {
```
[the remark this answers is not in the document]
ocdbytes: https://github.com/madara-alliance/madara-orchestrator/pull/101#discussion_r1741685821

e2e-tests/src/lib.rs
```diff
@@ -26,3 +45,158 @@ fn get_repository_root() -> PathBuf {
...
+    client
+}
+
+pub fn get_env_var(key: &str) -> Result<String, VarError> {
+    std::env::var(key)
+}
+
+pub fn get_env_var_or_panic(key: &str) -> String {
+    get_env_var(key).unwrap_or_else(|e| panic!("Failed to get env var {}: {}", key, e))
+}
```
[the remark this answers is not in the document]
ocdbytes: done

e2e-tests/src/localstack.rs
```diff
@@ -0,0 +1,219 @@
...
+use orchestrator::queue::job_queue::{
+    WorkerTriggerMessage, WorkerTriggerType, JOB_HANDLE_FAILURE_QUEUE, JOB_PROCESSING_QUEUE, JOB_VERIFICATION_QUEUE,
+    WORKER_TRIGGER_QUEUE,
+};
+use std::collections::HashMap;
+use std::fs::read;
+
+/// LocalStack struct
+pub struct LocalStack {
+    l2_block_number: String,
```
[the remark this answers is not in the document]
ocdbytes: I think it should be dependent on the block we are testing. renaming it folder_name will just cause confusion.

e2e-tests/src/mock_server.rs
ocdbytes: Yeah this code is not clear. So basically we have two. conditions to be inserted in URL so that's why I made this function need to replace this with correct implementation.

e2e-tests/src/mongodb.rs
ocdbytes: ok

e2e-tests/src/starknet_client.rs
ocdbytes: https://github.com/madara-alliance/madara-orchestrator/pull/101#discussion_r1741727936

e2e-tests/tests.rs
ocdbytes: So my intention to keep it there was in case we need to test some other block pie in e2e test.

e2e-tests/tests.rs
```diff
@@ -0,0 +1,164 @@
...
+    // Mocking the endpoint
+    mock_proving_job_endpoint_output(&mut sharp_client_mock).await;
+
+    // Step 3: DA job =================================================
+    // mocking get_block_call from starknet client
+
+    // Adding a mock da job so that worker does not create 60k+ jobs
+    put_job_data_in_db_da(&mongo_db_instance, l2_block_number.clone()).await;
+    mock_starknet_get_state_update(&mut starknet_client_mock, l2_block_number.clone()).await;
+    mock_starknet_get_nonce(&mut starknet_client_mock, l2_block_number.clone()).await;
```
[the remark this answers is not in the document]
ocdbytes: done

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -47,20 +46,47 @@ use {alloy::providers::RootProvider, alloy::transports::http::Http, reqwest::Cli
...
+/// until it finds a directory that contains both a 'Cargo.toml' file and a 'crates' directory.
+///
+/// # Returns
+///
+/// Returns a `PathBuf` representing the absolute path to the project root.
+///
+/// # Panics
+///
+/// Panics if it can't find a directory matching the criteria for the project root.
+pub fn find_project_root() -> PathBuf {
```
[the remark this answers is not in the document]
ocdbytes: which environment variable to use here because If I use `CARGO_MANIFEST_DIR` this will return the wrong directory that messes up with the path.

.env.test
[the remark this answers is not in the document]
apoorvsadana: the `SETTLEMENT_RPC_URL` is more generic though as we will reuse the same thing for L3s

e2e-tests/src/ethereum.rs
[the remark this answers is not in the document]
apoorvsadana: is this not possible?

crates/orchestrator/src/queue/job_queue.rs
apoorvsadana: do you remember what the error was? maybe we can add it as a comment

crates/prover-services/sharp-service/src/client.rs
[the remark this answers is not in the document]
apoorvsadana: accha, do you remember what the issue was? because technically we've the test cases for this as well which were working with `.json`

crates/settlement-clients/ethereum/src/lib.rs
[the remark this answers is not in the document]
apoorvsadana: i see, maybe we should just hardcode this as `${CARGO_MANIFEST_DIR}/../../`. logic to find root seems like an overkill 😅

e2e-tests/tests.rs
```diff
@@ -0,0 +1,317 @@
...
+) -> color_eyre::Result<Option<JobItem>> {
+    let mongo_db_client = get_mongo_db_client(mongo_db_server).await;
+    let collection = mongo_db_client.database("orchestrator").collection::<JobItem>("jobs");
+    let filter = doc! { "internal_id": l2_block_for_testing, "job_type" : "StateTransition" };
+    Ok(collection.find_one(filter, None).await.unwrap())
+}
+
+/// To get env variables to be used in testing
+fn get_env_vec() -> Vec<(String, String)> {
+    let env_vec = vec![
```
apoorvsadana, line 174: we can remove this as discussed

e2e-tests/src/localstack.rs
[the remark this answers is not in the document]
apoorvsadana: but there might be other e2e tests that aren't related to block numbers but need localstack, what would you do over here for that?

e2e-tests/src/sharp.rs
apoorvsadana: but why will the code change here? the test case will just call `add_mock_on_endpoint` wherever needed?

or let's say there are some common functions that all `Sharp` servers need to have, then the Sharp struct should contain a `mock_server` struct so that you directly import `add_mock_on_endpoint` and other mock server functionality. And then you can implement the custom stuff on top of `Shar`

e2e-tests/tests.rs
apoorvsadana: i see, but then we can change the number in the test case itself. the env var feels sort of incorrect here

e2e-tests/src/ethereum.rs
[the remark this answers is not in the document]
ocdbytes: Yes we can do this also . Will update the code for this.

crates/orchestrator/src/queue/job_queue.rs
ocdbytes: unable to serialise something error was there.

crates/prover-services/sharp-service/src/client.rs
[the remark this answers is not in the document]
ocdbytes: fixed

e2e-tests/tests.rs
```diff
@@ -0,0 +1,310 @@
...
+
+    println!("✅ Orchestrator setup completed.");
+
+    // Run orchestrator
+    let mut orchestrator = Orchestrator::run(setup_config.envs());
+    orchestrator.wait_till_started().await;
+
+    // Adding a case here to check for required state of the orchestrator to end the test.
+    let test_result =
+        wait_for_db_state(Duration::from_secs(1200), l2_block_number, setup_config.mongo_db_instance()).await;
```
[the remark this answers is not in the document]
ocdbytes: Cool What intervals to keep for checking expected states ?

e2e-tests/src/sharp.rs
ocdbytes: discussed

antiyro (approved):
lgtm

apoorvsadana (approved):

PR #103, opened 2024-08-31
Fix : generate merkle root function during fact calculation

ocdbytes:
- Previous behaviour : 
   When we compared the logic for the fact generation code for some other issue we found out that the logic of 
   `generate_merkle_root` function is not correct as per the cairo-lang repo code : 
    [generate_merkle_root()](https://github.com/starkware-libs/cairo-lang/blob/a86e92bfde9c171c0856d7b46580c66e004922f3/src/starkware/cairo/bootloaders/compute_fact.py#L47)

- New Behaviour : 
   Now the code is fixed and a simple test case is also added in order to check the logic compared to the code in starknet's cairo lang repo.

### DIFF
--- a/crates/prover-services/gps-fact-checker/src/fact_node.rs
+++ b/crates/prover-services/gps-fact-checker/src/fact_node.rs
@@ -24,6 +24,8 @@
 use alloy::primitives::{keccak256, B256};
 use cairo_vm::Felt252;
 use itertools::Itertools;
+use num_bigint::BigUint;
+use std::ops::Add;
 use utils::ensure;
 
 use super::error::FactCheckerError;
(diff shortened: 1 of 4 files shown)


crates/prover-services/gps-fact-checker/src/fact_node.rs
```diff
@@ -86,32 +88,88 @@ pub fn generate_merkle_root(
...
+
+/// Calculates the keccak hash and adds 1 to it.
+fn calculate_node_hash(node_data: &[u8]) -> B256 {
+    let hash = keccak256(node_data);
+    let hash_biguint = BigUint::from_bytes_be(hash.as_slice());
+    let incremented_hash = hash_biguint.add(BigUint::from(1u8));
+    let mut hash_bytes = incremented_hash.to_bytes_be();
+    while hash_bytes.len() < 32 {
+        hash_bytes.insert(0, 0);
+    }
```
apoorvsadana, line 129:
```
let mut padded_bytes = vec![0; 32 - hash_bytes.len()];
padded_bytes.append(&mut hash_bytes);
```

crates/prover-services/gps-fact-checker/src/fact_node.rs
```diff
@@ -86,32 +88,88 @@ pub fn generate_merkle_root(
...
+    /// `generate_output_root` function in cairo-lang repo.
+    /// We are comparing the output hash of the `generate_merkle_root` function
+    /// with our python output.
+    ///
+    /// Function link : https://github.com/starkware-libs/cairo-lang/blob/a86e92bfde9c171c0856d7b46580c66e004922f3/src/starkware/cairo/bootloaders/compute_fact.py#L47
+    ///
+    /// This will ensure that our logic here is correct.
+    #[test]
+    fn test_generate_merkle_root() {
+        let program_output_vec: Vec<Felt252> = vec![
```
apoorvsadana, line 151:
```rust
let program_output_vec: Vec<Felt252> = (1..=12).into_iter().map(|i| i.into()).collect();
```

odesenfans (approved):

Issue #17, opened 2024-06-17
feat: move important contract addresses to Config


apoorvsadana:
Running the orchestrator requires knowledge about multiple contract addresses - core contracts, verifier, memory page registry etc. We need to add these in the config

PR #106, opened 2024-09-03
Enhancement : Refactor Settings

ocdbytes:
- Issue : #17


crates/da-clients/ethereum/src/config.rs
```diff
@@ -1,34 +1,20 @@
...
-
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct EthereumDaConfig {
     pub rpc_url: String,
     pub memory_pages_contract: String,
     pub private_key: String,
 }
 
-#[async_trait]
```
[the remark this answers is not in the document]
ocdbytes: What is the reason for this to be async ? this is config which should not be an async operation.

crates/da-clients/ethereum/src/lib.rs
```diff
@@ -37,3 +45,14 @@ impl DaClient for EthereumDaClient {
         131072
     }
 }
+
+impl EthereumDaClient {
+    pub fn new_with_settings(settings: &impl Settings) -> Self {
```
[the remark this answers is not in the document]
ocdbytes: https://github.com/madara-alliance/madara-orchestrator/pull/106#discussion_r1746866382

crates/orchestrator/src/alerts/aws_sns/mod.rs
```diff
@@ -1,27 +1,32 @@
...
 use aws_sdk_sns::config::Region;
 use aws_sdk_sns::Client;
-use utils::env_utils::get_env_var_or_panic;
+use utils::settings::Settings;
+
+pub const AWS_SNS_SETTINGS_NAME: &str = "sns";
 
 pub struct AWSSNS {
     client: Client,
+    topic_arn: String,
```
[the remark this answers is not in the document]
ocdbytes: I need to discuss this with you. I am not in favour of passing aws_config through here : 
https://github.com/madara-alliance/madara-orchestrator/pull/107/files#diff-cbe058cb32b3323389374453c6b2db7d52432aacf1a7e59d3ee2ee99364bccc4R187

crates/orchestrator/src/data_storage/aws_s3/mod.rs
```diff
@@ -20,11 +22,13 @@ pub struct AWSS3 {
 ///
 /// - initializing a new AWS S3 client
 impl AWSS3 {
-    /// Initializes a new AWS S3 client by passing the config
-    /// and returning it.
-    pub fn new(s3_config: AWSS3Config, aws_config: &SdkConfig) -> Self {
+    /// To init the struct with main settings
+    pub async fn new_with_settings(settings: &impl Settings) -> Self {
+        let s3_config = AWSS3Config::new_with_settings(settings);
+        let aws_config = aws_config::load_from_env().await;
```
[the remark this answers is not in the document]
ocdbytes: we are using settings provider and we made the functions assuming that some day some other providers will also be implemented. Idts it is ideal to pass a config of certain provider as an argument to a generalised function. We can discuss it because I don't agree with this design or Ig I am missing something.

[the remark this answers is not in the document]
apoorvsadana: You're right but what am I saying is that we don't build the aws config by directly reading the env or something. We build it using the settings provider. And then we pass that AWS config downwards, this way 

1. We're still agnostic to what provider is being used
2. We reduce duplication by creating the AWS config only once

crates/da-clients/ethereum/src/config.rs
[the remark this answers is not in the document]
apoorvsadana: in celestia it wasn't possible to do it sync, can you check with @heemankv once?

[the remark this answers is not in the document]
ocdbytes: let me check

crates/orchestrator/src/config.rs
```diff
@@ -49,6 +49,31 @@ pub struct Config {
     alerts: Box<dyn Alerts>,
 }
 
+/// `ProviderConfig` is an enum used to represent the global config built
+/// using the settings provider. More providers can be added eg : GCP, AZURE etc.
+pub enum ProviderConfig {
+    AWS(Box<SdkConfig>),
+    INVALID,
```
apoorvsadana, line 56: why do we need this?

crates/orchestrator/src/config.rs
```diff
@@ -49,6 +49,31 @@ pub struct Config {
     alerts: Box<dyn Alerts>,
 }
 
+/// `ProviderConfig` is an enum used to represent the global config built
+/// using the settings provider. More providers can be added eg : GCP, AZURE etc.
+pub enum ProviderConfig {
+    AWS(Box<SdkConfig>),
```
apoorvsadana, line 55: why box here?

crates/orchestrator/src/config.rs
```diff
@@ -58,27 +83,23 @@ pub async fn init_config() -> Config {
...
 
-    // init AWS
-    let aws_config = aws_config::load_from_env().await;
+    // init database
+    let database = build_database_client(&settings_provider).await;
+    let da_client = build_da_client(&settings_provider).await;
+    let settlement_client = build_settlement_client(&settings_provider).await;
+    let prover_client = build_prover_service(&settings_provider);
+    let storage_client =
+        build_storage_client(&settings_provider, ProviderConfig::AWS(Box::new(aws_config.clone()))).await;
```
apoorvsadana, line 95: we could use Arc<ProviderConfig>

apoorvsadana (approved):

crates/orchestrator/src/config.rs
ocdbytes: clippy. error aa rha tha

ocdbytes: Sdk config ka size kaafi bda h.

crates/orchestrator/src/config.rs
ocdbytes: abhi code me jha pr me match kr rha hu providerconfig ko vha pr I need to handle the other cases with a panic. so when it was alone AWS it was giving code unreachable error. So that's why added this enum temporarily.

akhercha (approved):
LGTM!

crates/orchestrator/src/config.rs
apoorvsadana: intersting, can you comment that?

crates/orchestrator/src/config.rs
apoorvsadana: that's not the correct design. you don't need add another `enum` so that you can panic. if there was only AWS over here, the user couldn't pass anything else, the code won't compile so there's no issue

PR #109, opened 2024-09-04
Feat : Database/timestamp

ocdbytes:
- Added created_at and updated_at timestamp in job document in database.
- Issue : #102

### DIFF
--- a/crates/orchestrator/Cargo.toml
+++ b/crates/orchestrator/Cargo.toml
@@ -27,6 +27,7 @@ bincode = { workspace = true }
 bytes = "1.6.0"
 c-kzg = { workspace = true }
 cairo-vm = { workspace = true }
+chrono = "0.4.0"
 color-eyre = { workspace = true }
 da-client-interface = { workspace = true }
 dotenvy = { workspace = true }
@@ -63,7 +64,6 @@ tracing-subscriber = { workspace = true, features = ["env-filter"] }
 url = { workspace = true }
 utils = { workspace = true }
 uuid = { workspace = true, features = ["v4", "serde"] }
-chrono = "0.4.0"
 
 [features]
 default = ["ethereum", "with_mongodb", "with_sqs"]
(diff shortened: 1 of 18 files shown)


crates/orchestrator/Cargo.toml
```diff
@@ -63,6 +63,7 @@ tracing-subscriber = { workspace = true, features = ["env-filter"] }
 url = { workspace = true }
 utils = { workspace = true }
 uuid = { workspace = true, features = ["v4", "serde"] }
+chrono = "0.4.0"
```
apoorvsadana: should come from workspace?

crates/orchestrator/src/database/mongodb/mod.rs
```diff
@@ -64,17 +65,20 @@ impl MongoDb {
         if result.modified_count == 0 {
             return Err(eyre!("Failed to update job. Job version is likely outdated"));
         }
-        self.update_job_version(current_job).await?;
+        self.post_job_update(current_job).await?;
```
apoorvsadana, line 68: i can't recall why we were doing this actually? why don't we just increment the version inside `update`?

ocdbytes: We actually discussed about this previously. Some mongo issue was there when injecting this inc update in the actual document for updating. That's why at that time we agreed for this approach.

crates/orchestrator/Cargo.toml
ocdbytes: updated

crates/orchestrator/src/database/mongodb/mod.rs
apoorvsadana: Ok, let's create an issue for this and try to tackle it. Because even if we go down this route then I think the two updates should be part of a txn

akhercha (approved):
Looks good!

Issue #99, opened 2024-08-23
Workspace oriented dependencies.


heemankv:
Define all the dependencies at the workspace level and then pull them in each crate.
https://github.com/madara-alliance/madara-orchestrator/pull/81#discussion_r1725067804

heemankv:
Going with this convention for PR : 
- If a dependencies is only being used in a single crate and it's specific to that crate then no need to move it to workspace dependency.
- If a dependency is being used in multiple crates but with different features then it should be defined as a workspace dependency and individual features must be pulled in each crate.
- If any dependency is being used in many crates without any `ifs and buts` then we define it as a workspace dependency.

PR #107, opened 2024-09-04
Feature/drilled config

heemankv:
# This PR resolves #104.

### Introduces : 
- Config drilled from `parent` to `child` functions, making it more pure and removing dirty access of global config.
- New `TestConfigBuilder` with sync `config` method and only `build` async method.

PR #115, opened 2024-09-06
workspace oriented dependencies

heemankv:
# This PR resolves issue #99

Going with this convention for PR : 
- If a dependencies is only being used in a single crate and it's specific to that crate then no need to move it to workspace dependency.
- If a dependency is being used in multiple crates but with different features then it should be defined as a workspace dependency and individual features must be pulled in each crate.
- If any dependency is being used in many crates without any `ifs and buts` then we define it as a workspace dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,9 +34,13 @@ dotenvy = "0.15.7"
 futures = "0.3.30"
 mongodb = { version = "2.8.1" }
 omniqueue = { version = "0.2.0" }
-reqwest = { version = "0.11.24", features = ["rustls-tls", "native-tls"] }
+reqwest = { version = "0.12.7", features = [
+  "rustls-tls",
+  "native-tls",
+  "json",
+] }
 rstest = "0.18.2"
-serde = { version = "1.0.197" }
+serde = { version = "1.0.197", features = ["derive"] }
 serde_json = "1.0.114"
 starknet = "0.9.0"
 tempfile = "3.8.1"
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -50,13 +54,13 @@ url = { version = "2.5.0", features = ["serde"] }
 uuid = { version = "1.7.0", features = ["v4", "serde"] }
 httpmock = { version = "0.7.0", features = ["remote"] }
 num-bigint = { version = "0.4.4" }
-arc-swap = { version = "1.7.1" }
+arc-swap = { version = "1.7.0" }
```
heemankv, line 57: `arc-swap` should be removed after PR #107  get's merged.

crates/settlement-clients/ethereum/Cargo.toml
```diff
@@ -4,18 +4,18 @@ version.workspace = true
 edition.workspace = true
 
 [dependencies]
-alloy = { workspace = true, features = ["full", "node-bindings"] }
 alloy-primitives = { version = "0.7.7", default-features = false }
+
+alloy = { workspace = true, features = ["full", "node-bindings"] }
```
heemankv, line 9: Wanted to update alloy to latest version, but it's specifically required with varied features in the crates. Left it as it is for now.
Screenshot 2024-09-06 at 12 38 53 PM

heemankv:
Completed Self Review, looks good!

crates/settlement-clients/ethereum/Cargo.toml
apoorvsadana: are you sure, it's possible that we just copied the import from somewhere and maybe the both work on the latest version?

crates/da-clients/ethereum/Cargo.toml
```diff
@@ -15,14 +15,14 @@ alloy = { git = "https://github.com/alloy-rs/alloy", rev = "68952c0", features =
...
 ] }
 async-trait = { workspace = true }
-c-kzg = "1.0.0"
+c-kzg = { workspace = true }
 color-eyre = { workspace = true }
 da-client-interface = { workspace = true }
-dotenv = "0.15"
-mockall = "0.12.1"
-reqwest = { version = "0.12.3" }
+dotenvy.workspace = true
```
apoorvsadana, line 21: Nitpick: should we change everything to one format? either `{ workspace = true }` or `.workspace = true`?

apoorvsadana (approved):

PR #116, opened 2024-09-06
Chore: Rust Toolchain

heemankv:
- Rust toolchain defined as 1.81.0.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "stable"
+channel = "1.81.0"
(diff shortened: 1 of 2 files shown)


rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "stable"
+channel = "1.81.0"
```
apoorvsadana, line 2: can we also add nightly?

heemankv: Issue being reassigned to PR #105.

heemankv:
Issue being reassigned to PR #105

PR #122, opened 2024-09-13
feat: Snos job

akhercha:
Resolves: #11 

Follow-up of: https://github.com/madara-alliance/madara-orchestrator/pull/105

## Added

- Implementation of the `SnosJob`.

## Updated
- toolchain to nightly so it matches the one used in the snos repository. It caused a lot of lint updates, hence all the files updated in this PR,
- some package versions so they're compatible with snos.


apoorvsadana:
This looks good. The e2e seems to fail, @ocdbytes is looking into it and he will create a PR to add SNOS run in the e2e. After that we can merge this :)

apoorvsadana:
Closing in favor of #136

PR #125, opened 2024-09-18
move starknet settlement to zaun

byteZorvin:
- Orchestrator now uses [zaun](https://github.com/keep-starknet-strange/zaun) types now to interact with starknet core contract.


crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -27,10 +27,10 @@ use mockall::{automock, predicate::*};
 
 use alloy::providers::ProviderBuilder;
 use conversion::{get_input_data_for_eip_4844, prepare_sidecar};
-#[cfg(feature = "testing")]
-use settlement_client_interface::{SettlementClient, SettlementConfig, SettlementVerificationStatus};
 #[cfg(not(feature = "testing"))]
-use settlement_client_interface::{SettlementClient, SettlementConfig, SettlementVerificationStatus};
+use settlement_client_interface::{SettlementClient, SettlementVerificationStatus};
+#[cfg(feature = "testing")]
+use settlement_client_interface::{SettlementClient, SettlementVerificationStatus};
```
apoorvsadana, line 33: what's the different in testing and not testing?

crates/settlement-clients/starknet/Cargo.toml
```diff
@@ -4,19 +4,26 @@ version.workspace = true
 edition.workspace = true
 
 [dependencies]
+anyhow = "1.0.89"
+appchain-core-contract-client = { workspace = true }
 async-trait = { workspace = true }
 c-kzg = { workspace = true }
 color-eyre = { workspace = true }
+dotenvy.workspace = true
+env_logger = "0.11.5"
```
apoorvsadana, line 14: so do we need `env_logger` finally? also lets import common crates to the global `Cargo.toml` and then import here

crates/settlement-clients/starknet/src/lib.rs
```diff
@@ -106,44 +117,42 @@ impl SettlementClient for StarknetSettlementClient {
...
-            MaybePendingTransactionReceipt::PendingReceipt(tx) => match tx.execution_result() {
-                ExecutionResult::Succeeded => Ok(SettlementVerificationStatus::Pending),
-                ExecutionResult::Reverted { reason } => Ok(SettlementVerificationStatus::Rejected(format!(
+        let execution_result = tx_receipt.receipt.execution_result();
+        let status = execution_result.status();
+
+        if tx_receipt.block.is_pending() {
+            match status {
+                TransactionExecutionStatus::Succeeded => Ok(SettlementVerificationStatus::Pending),
+                TransactionExecutionStatus::Reverted => Ok(SettlementVerificationStatus::Rejected(format!(
```
apoorvsadana, line 142: technically we can do the reverted check before checking for `is_pending` right? will reduce lines of code

crates/settlement-clients/starknet/src/lib.rs
```diff
@@ -197,11 +206,26 @@ impl SettlementClient for StarknetSettlementClient {
...
             return Err(eyre!("Could not fetch last block number from core contract."));
         }
-        Ok(block_number[0].try_into()?)
+
+        Ok(u64::from_le_bytes(block_number[0].to_bytes_le()[0..8].try_into().unwrap()))
     }
 
     /// Returns the nonce for the wallet in use.
     async fn get_nonce(&self) -> Result<u64> {
         todo!("Yet to impl nonce call for Starknet.")
```
apoorvsadana, line 216: can you implement this as well

crates/settlement-clients/starknet/src/lib.rs
```diff
@@ -197,11 +206,26 @@ impl SettlementClient for StarknetSettlementClient {
         if block_number.is_empty() {
             return Err(eyre!("Could not fetch last block number from core contract."));
         }
-        Ok(block_number[0].try_into()?)
+
+        Ok(u64::from_le_bytes(block_number[0].to_bytes_le()[0..8].try_into().unwrap()))
```
apoorvsadana, line 211: 1. can you move this to a function in `conversions.rs`
2. the test case should call this function and the code as well
3. the function should throw an error if bytes from 8-32 are non zero because then the conversion will be invalid
4. nitpick: i have usually seen `be` being used more commonly and and it's easier to understand because that's how we write stuff as well, should we use that?

crates/settlement-clients/starknet/src/tests/mock_contracts/.gitignore
apoorvsadana, line 1: should we add these to the global gitignore?

crates/settlement-clients/starknet/src/tests/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+    collections::HashMap,
+    future::Future,
+    path::{Path, PathBuf},
+    process::{Child, Command, Output, Stdio},
+    str::FromStr,
+    time::Duration,
+};
+use tempfile::TempDir;
+
+pub struct MadaraCmd {
```
apoorvsadana, line 18: let's add a comment pointing to where we took the code from

crates/settlement-clients/starknet/src/tests/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+        if attempt >= 10 {
+            panic!("No answer from the node after {attempt} attempts: {:#}", err)
+        }
+
+        tokio::time::sleep(duration).await;
+    }
+}
+
+impl MadaraCmd {
+    pub fn wait_with_output(mut self) -> Output {
```
apoorvsadana, line 44: we can remove this if we aren't using it

crates/settlement-clients/starknet/src/tests/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+        self.process.take().unwrap().wait_with_output().unwrap()
+    }
+
+    pub fn json_rpc(&mut self) -> &JsonRpcClient<HttpTransport> {
```
apoorvsadana, line 48: same for this

crates/settlement-clients/starknet/src/tests/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+                anyhow::Ok(())
+            },
+            Duration::from_millis(1000),
+        )
+        .await;
+        self.ready = true;
+        self
+    }
+
+    pub async fn wait_for_sync_to(&mut self, block_n: u64) -> &mut Self {
```
apoorvsadana, line 71: we probably don't need this as well as it's for full node

crates/settlement-clients/starknet/src/tests/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+        self
+    }
+
+    pub fn env(mut self, env: impl IntoIterator<Item = (impl Into<String>, impl Into<String>)>) -> Self {
+        self.env = env.into_iter().map(|(k, v)| (k.into(), v.into())).collect();
+        self
+    }
+
+    pub fn run(self) -> MadaraCmd {
+        let target_bin = option_env!("COVERAGE_BIN").unwrap_or("./target/debug/madara");
```
apoorvsadana, line 170: let's just do an `expect` instead of a default value? will there be any case where the madara binary will be in `./target/debug/madara`

crates/settlement-clients/starknet/src/tests/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+#[fixture]
+pub fn set_workdir() {
+    dotenvy::from_filename_override(".env.test").expect("Failed to load the .env file");
+    env::set_current_dir(env::var("MADARA_BINARY_PATH").unwrap()).expect("Failed to set working directory");
+}
+
+#[ignore]
+#[rstest]
+#[tokio::test]
+async fn madara_can_sync_a_few_blocks(_set_workdir: ()) {
```
apoorvsadana, line 213: we can remove this

crates/settlement-clients/starknet/src/tests/mod.rs
apoorvsadana, line 1: we can probably move this code into `setup.rs` or `utils.rs`?

crates/settlement-clients/starknet/src/tests/test.rs
```diff
@@ -0,0 +1,229 @@
...
+            "--chain-config-override",
+            "--block-time",
+            "1",
+        ])
+        .run();
+    node.wait_for_ready().await;
+    node
+}
+
+async fn wait_for_tx_success(
```
apoorvsadana, line 48: can we actually reuse `wait_for_cond` function here?

crates/settlement-clients/starknet/src/tests/test.rs
```diff
@@ -0,0 +1,229 @@
...
+                TransactionExecutionStatus::Succeeded => return true,
+                TransactionExecutionStatus::Reverted => return false,
+            },
+            TransactionStatus::AcceptedOnL1(status) => match status {
+                TransactionExecutionStatus::Succeeded => return true,
+                TransactionExecutionStatus::Reverted => return false,
+            },
+        }
+        // This is done, since currently madara does not increment nonce for pending transactions
+        tokio::time::sleep(duration).await;
```
apoorvsadana, line 81: instead of this, should we wait for block to be not pending?

crates/settlement-clients/starknet/src/tests/test.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+#[fixture]
+async fn setup() -> (SingleOwnerAccount<JsonRpcClient<HttpTransport>, LocalWallet>, MadaraCmd) {
+    let _ = env_logger::builder().is_test(true).try_init();
+    dotenvy::from_filename_override(".env.test").expect("Failed to load the .env file");
+
+    let madara_process = spin_up_madara().await;
+    env::set_var("STARKNET_RPC_URL", madara_process.rpc_url.to_string());
+    println!("RPC url {:?}", madara_process.rpc_url);
```
apoorvsadana, line 92:
```suggestion
```

crates/settlement-clients/starknet/src/tests/test.rs
```diff
@@ -0,0 +1,229 @@
...
+};
+use std::env;
+use std::path::Path;
+use std::sync::Arc;
+use std::time::Duration;
+use utils::settings::env::EnvSettingsProvider;
+use utils::settings::Settings;
+
+pub async fn spin_up_madara() -> MadaraCmd {
+    println!("Spinning up madara");
```
apoorvsadana, line 25: instad of println! we can maybe use `log::trace` or remove it

crates/settlement-clients/starknet/src/tests/test.rs
```diff
@@ -0,0 +1,229 @@
...
+    dotenvy::from_filename_override(".env.test").expect("Failed to load the .env file");
+
+    let madara_process = spin_up_madara().await;
+    env::set_var("STARKNET_RPC_URL", madara_process.rpc_url.to_string());
+    println!("RPC url {:?}", madara_process.rpc_url);
+
+    let env_settings = EnvSettingsProvider::default();
+    let rpc_url = Url::parse(&env_settings.get_settings_or_panic("STARKNET_RPC_URL")).unwrap();
+
+    // let provider = JsonRpcClient::new(HttpTransport::new(Url::parse("http://localhost:9944").unwrap()));
```
apoorvsadana, line 97:
```suggestion
```

crates/settlement-clients/starknet/src/tests/test.rs
```diff
@@ -0,0 +1,229 @@
...
+        tokio::time::sleep(duration).await;
+    }
+}
+
+#[fixture]
+async fn setup() -> (SingleOwnerAccount<JsonRpcClient<HttpTransport>, LocalWallet>, MadaraCmd) {
+    let _ = env_logger::builder().is_test(true).try_init();
+    dotenvy::from_filename_override(".env.test").expect("Failed to load the .env file");
+
+    let madara_process = spin_up_madara().await;
```
apoorvsadana, line 90: should we make `spin_up_madara` a fixture as well?

crates/settlement-clients/starknet/src/tests/test.rs
```diff
@@ -0,0 +1,229 @@
...
+
+    // `SingleOwnerAccount` defaults to checking nonce and estimating fees against the latest
+    // block. Optionally change the target block to pending with the following line:
+    account.set_block_id(BlockId::Tag(BlockTag::Pending));
+    (account, madara_process)
+}
+
+#[rstest]
+#[tokio::test]
+async fn test_deployment(#[future] setup: (SingleOwnerAccount<JsonRpcClient<HttpTransport>, LocalWallet>, MadaraCmd)) {
```
apoorvsadana, line 115: can you create a type alias for `SingleOwnerAccount<JsonRpcClient<HttpTransport>, LocalWallet>` for easier readability

crates/settlement-clients/starknet/src/tests/test.rs
apoorvsadana, line 48:
```suggestion
async fn wait_for_tx(
```

crates/settlement-clients/starknet/src/tests/test.rs
```diff
@@ -0,0 +1,229 @@
...
+    let InvokeTransactionResult { transaction_hash: deploy_tx_hash } =
+        deploy_v1.send().await.expect("Unable to deploy contract");
+
+    let is_success = wait_for_tx_success(&account, deploy_tx_hash, Duration::from_secs(2)).await;
+    assert!(is_success, "Deploy trasaction failed");
+}
+
+#[rstest]
+#[tokio::test]
+async fn test_settle(#[future] setup: (SingleOwnerAccount<JsonRpcClient<HttpTransport>, LocalWallet>, MadaraCmd)) {
```
apoorvsadana, line 157: we can just keep this test case and remove `test_deployment` right?

crates/settlement-clients/starknet/src/tests/test.rs
```diff
@@ -0,0 +1,229 @@
...
+    let settlement_client = StarknetSettlementClient::new_with_settings(&env_settings).await;
+    let onchain_data_hash = [1; 32];
+    let mut program_output = Vec::with_capacity(32);
+    program_output.fill(onchain_data_hash);
+    let update_state_tx_hash = settlement_client
+        .update_state_calldata(program_output, onchain_data_hash, [1, 1])
+        .await
+        .expect("Sending Update state");
+
+    println!("update state tx hash {:?}", update_state_tx_hash);
```
apoorvsadana, line 206: println! to logs

crates/settlement-clients/starknet/src/tests/test.rs
```diff
@@ -0,0 +1,229 @@
...
+            FunctionCall {
+                contract_address: deployed_address,
+                entry_point_selector: selector!("get_is_updated"),
+                calldata: vec![Felt::from_bytes_be_slice(&onchain_data_hash)],
+            },
+            BlockId::Tag(BlockTag::Latest),
+        )
+        .await
+        .expect("failed to call the contract");
+    assert!(call_result[0] == true.into(), "Should be updated");
```
apoorvsadana, line 228: doesn't the call return the actual hash?

.github/workflows/starknet-test.yml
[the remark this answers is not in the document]
byteZorvin: moved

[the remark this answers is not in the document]
byteZorvin: moved

.gitignore
```diff
@@ -5,7 +5,9 @@
 
 *.code-workspace
 .vscode
+db/
```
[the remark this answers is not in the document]
byteZorvin: Yes correct, removed.

crates/settlement-clients/ethereum/src/lib.rs
byteZorvin: yes this was a mistake, removed extra code.

crates/settlement-clients/starknet/src/tests/test.rs
byteZorvin: Call does not create a starknet transaction, so it does not have a hash

.env.example
```diff
@@ -47,4 +47,12 @@ MADARA_RPC_URL=
...
 PRIVATE_KEY=
 ETHEREUM_PRIVATE_KEY=
-STARKNET_SOLIDITY_CORE_CONTRACT_ADDRESS=
\ No newline at end of file
+STARKNET_SOLIDITY_CORE_CONTRACT_ADDRESS=
+
+
+
+##### STARKNET SETTLEMENT TEST #####
+STARKNET_PRIVATE_KEY=
```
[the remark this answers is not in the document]
byteZorvin: 1. Removed `STARKNET_RPC_URL` from the .env.test
2. changed public_key to address
3. Makes sense, removed the key from `.env.example`

crates/settlement-clients/starknet/src/tests/mod.rs
byteZorvin: That would be better, moved !

crates/settlement-clients/starknet/src/lib.rs
byteZorvin: 1. Created a new function
2. Added a testcase for the conversion as well
3. the functions asserts if the value is within `u64` range and bytes from 8 to 31 are zero
4. using `be` now

crates/settlement-clients/starknet/src/tests/test.rs
byteZorvin: Removed

crates/settlement-clients/starknet/src/tests/test.rs
byteZorvin: Created

crates/settlement-clients/starknet/src/tests/test.rs
byteZorvin: Removed

crates/settlement-clients/starknet/src/lib.rs
byteZorvin: Updated this

crates/settlement-clients/starknet/src/lib.rs
byteZorvin: Implemented

crates/settlement-clients/starknet/src/tests/test.rs
byteZorvin: converted to fixture

crates/settlement-clients/starknet/src/tests/test.rs
byteZorvin: changed to logs

crates/settlement-clients/starknet/src/tests/test.rs
byteZorvin: I assume, this achieves the same, also I am not sure how does one go from transaction hash to block status

.env.example
```diff
@@ -47,4 +47,10 @@ MADARA_RPC_URL=
 MEMORY_PAGES_CONTRACT_ADDRESS=
 PRIVATE_KEY=
 ETHEREUM_PRIVATE_KEY=
-STARKNET_SOLIDITY_CORE_CONTRACT_ADDRESS=
\ No newline at end of file
+STARKNET_SOLIDITY_CORE_CONTRACT_ADDRESS=
+
+
+
+##### STARKNET SETTLEMENT TEST #####
```
apoorvsadana, line 54:
```suggestion
##### STARKNET SETTLEMENT (L3s) #####
```

.github/workflows/coverage.yml
```diff
@@ -62,14 +66,42 @@ jobs:
...
+        run: |
+          cd crates/settlement-clients/starknet/src/tests/mock_contracts
+          scarb fmt --check
+          scarb build
+
+      - name: Checkout Madara
+        uses: actions/checkout@v4
+        with:
+          repository: madara-alliance/madara # Replace with your first repository
+          ref: "0c821aecd1cafc516d00548473fcaddaa4dc0ce4" # Replace with your branch name
```
apoorvsadana, line 84:
```suggestion
          repository: madara-alliance/madara
          ref: "0c821aecd1cafc516d00548473fcaddaa4dc0ce4"
```

crates/settlement-clients/starknet/Cargo.toml
apoorvsadana: do we need this

crates/settlement-clients/starknet/src/lib.rs
apoorvsadana: i would ideally add a small test case for this :)

crates/settlement-clients/starknet/src/conversion.rs
```diff
@@ -1,9 +1,32 @@
...
     slices.iter().map(slice_u8_to_field).collect()
 }
 
-pub(crate) fn slice_u8_to_field(slice: &[u8; 32]) -> FieldElement {
-    FieldElement::from_byte_slice_be(slice).expect("could not convert u8 slice to FieldElement")
+pub(crate) fn slice_u8_to_field(slice: &[u8; 32]) -> Felt {
+    Felt::from_bytes_be_slice(slice)
+}
+
+pub(crate) fn u64_from_felt(number: Felt) -> u64 {
```
apoorvsadana, line 11: can we make this return a Result instead? if people call this normally in the code, they might not realise that this function can panic

crates/settlement-clients/starknet/src/lib.rs
```diff
@@ -197,11 +204,26 @@ impl SettlementClient for StarknetSettlementClient {
...
+        Ok(u64_from_felt(self.account.get_nonce().await?))
+    }
+}
+
+#[cfg(test)]
+mod test {
+
+    use starknet::core::types::Felt;
+    #[test]
+    fn test_felt_conversion() {
```
apoorvsadana, line 222: do we need this test here now?

crates/settlement-clients/starknet/src/tests/test.rs
apoorvsadana: pending?

crates/settlement-clients/starknet/src/tests/test.rs
apoorvsadana: there are some println! pending, can you check

crates/settlement-clients/starknet/src/tests/test.rs
apoorvsadana: oh right my bad

crates/settlement-clients/starknet/src/lib.rs
byteZorvin: yes removed

crates/settlement-clients/starknet/src/tests/test.rs
byteZorvin: No println logs now

PR #126, opened 2024-09-19
Using Self hosted GitHub CI runners

k4kratik:


apoorvsadana:
Closing this for now as self hosted runner don't have a lot of extra advantage for the orchestrator. Let's integrate it only for `madara` for now.

PR #133, opened 2024-09-30
change calldata type to

apoorvsadana:


byteZorvin (approved):
Lgtm, this should be better as u8 array seems much more standard as well.

Issue #112, opened 2024-09-05
Database update logic fix


ocdbytes:
- Current Behaviour : 
Current when we are updating the job using the mongo db `update_job_optimistically` function [code](https://github.com/madara-alliance/madara-orchestrator/blob/main/crates/orchestrator/src/database/mongodb/mod.rs#L57) Here we are using a `post_job_update` function to update the version and timestamp of the job. 

- Expected Behaviour : 
This should be happening in a single db transaction.

PavitraAgarwal21:
@ocdbytes can i take up this issue

ocdbytes:
@PavitraAgarwal21 

Unfortunately we are not taking external contributions right now.

PR #134, opened 2024-09-30
Betterment/update job

heemankv:
## This PR solves #112.

- Remove `update_job_status`, `update_metadata`, `update_job_optimistically`, `post_job_update` functions from `Database` implementation.
- Reimplemented `update_job` to use `JobItemUpdates`.
- Provide `JobItemUpdates` object with all the needed changes only, which get's reflected on the database.

### WHY this approach?
- Code had three functions `update_job`, `update_job_status` & `update_metadata` to update job values in database.
- `update_job` was being passed with the whole `jobItem` to update, in it's usage it was only changing 2-3 values in the job and not the whole job object, resulting in other values unnecessarily being rewritten. (for eg, same version value is being sent within jobItem, resulting in a pointless rewrite 1 -> 1).
- The above is also why the code had to update `version` and `updated_at` fields post these updates, since first the version get's rewritten (1->1) and then it's overwritten to an increased value in `post_job_update` (1->2).
- `update_job_status` and `update_metadata` followed the same trend but only updated individual values.

### WHAT now ?
- Trimming the above problem, reinventing `update_job` seemed the best approach to move forward with.
- Provide a `JobItemUpdates` object that consist of all the updates that needs to be done to the job in a single query.
- Code doesn't allow updating `id` and `created_at`. 
- Code implemented `only_status` and `only_metadata` to allow for cleaner code for these updates.
Screenshot 2024-09-30 at 4 49 48 PM
Screenshot 2024-09-30 at 4 51 09 PM


heemankv:
Done with Self Review,
I think we are missing a test case for the new `update_job`.
WDYT ? @apoorvsadana

apoorvsadana (commented):
This looks cleaner but I have two issues before the reivew

1. The version shouldn't come from the user. It's a invisible implementation that you don't need to worry about when writing the code. All you should know is that if your job is stale the update won't work.
2. If tomorrow I add a new field to `Job` and call `update_job`, it might feel that it would just work but actually I would need to update the logic and handle the if else. Is there a way to catch this silent failure later?

heemankv:
> 1. The version shouldn't come from the user. It's a invisible implementation that you don't need to worry about when writing the code. All you should know is that if your job is stale the update won't work.

Updated the logic to don't let parent fn to decide whether to update `version` or not, it will now always be updated.
Additionally, the same logic is now followed for `updated_at`.

crates/orchestrator/src/jobs/types.rs
```diff
@@ -138,6 +140,57 @@ pub struct JobItem {
...
+        JobItemUpdates { metadata: Some(metadata), ..Default::default() }
+    }
+    pub fn to_document(&self, current_job: &JobItem) -> Result<Document> {
+        let mut doc = Document::new();
+
+        // Serialize the struct to BSON
+        let bson = bson::to_bson(&self)?;
+
+        // If serialization was successful and it's a document
+        if let Bson::Document(bson_doc) = bson {
```
apoorvsadana, line 177: if this doesn't work should we throw an error?

crates/orchestrator/src/jobs/types.rs
```diff
@@ -138,6 +140,57 @@ pub struct JobItem {
...
+            // Add non-null fields to our document
+            for (key, value) in bson_doc.iter() {
+                if !matches!(value, Bson::Null) {
```
apoorvsadana, line 180: if everything is null should we also throw an error with nothing to update maybe?

crates/orchestrator/src/jobs/types.rs
apoorvsadana, line 177: this logic is specific to mongodb, how can we generalise the code so that this would also work for postgres down the line?

crates/orchestrator/src/jobs/mod.rs
```diff
@@ -344,10 +368,17 @@ pub async fn handle_job_failure(id: Uuid, config: Arc<Config>) -> Result<(), Job
     }
 
     metadata.insert("last_job_status".to_string(), job.status.to_string());
-    job.metadata = metadata;
-    job.status = JobStatus::Failed;
-
-    config.database().update_job(&job).await.map_err(|e| JobError::Other(OtherError(e)))?;
+    config
+        .database()
+        .update_job(&job, JobItemUpdates {
```
apoorvsadana, line 373: builder pattern?

crates/orchestrator/src/tests/database/mod.rs
```diff
@@ -158,50 +161,42 @@ async fn database_get_jobs_after_internal_id_by_job_type_works() {
...
+            external_id: None,
+            internal_id: None,
+            job_type: None,
+        })
+        .await;
+
+    let mut job_after_updates = build_job_item(JobType::DataSubmission, JobStatus::LockedForProcessing, 456);
+    job_after_updates.version += 1;
+
+    if let Some(job_after_updates_db) = database_client.get_job_by_id(job_id).await.unwrap() {
```
apoorvsadana, line 194: we should throw an error if it's none?

crates/orchestrator/src/tests/database/mod.rs
```diff
@@ -158,50 +161,42 @@ async fn database_get_jobs_after_internal_id_by_job_type_works() {
...
+        .update_job(&job_cloned, JobItemUpdates {
+            status: Some(JobStatus::LockedForProcessing),
+            metadata: Some(updated_metadata),
+            external_id: None,
+            internal_id: None,
+            job_type: None,
+        })
+        .await;
+
+    let mut job_after_updates = build_job_item(JobType::DataSubmission, JobStatus::LockedForProcessing, 456);
```
apoorvsadana, line 191: nitpick: it's cleaner if we edit the original job here instead of building a new one? or maybe we hardcode `asserts` to the new values and for old values we take them from `job`

crates/orchestrator/src/jobs/types.rs
heemankv: `
let bson = bson::to_bson(&self)?;`
using the ? we are already throwing the error into a `color_eyre` report 

`if let Bson::Document(bson_doc) = bson {`
Bson is an enum, off which Document is a embedded type,

What else is needed ?
What am I missing here ? @apoorvsadana 


Screenshot 2024-10-07 at 3 18 39 PM
Screenshot 2024-10-07 at 3 19 01 PM

crates/orchestrator/src/jobs/types.rs
heemankv: While writing this code I assumed such case would not happen, lemme introduce a flag that checks if the not null case was ever executed, this will let us know if anything is there to be updated or not.

heemankv: [Here is the impl](https://github.com/madara-alliance/madara-orchestrator/pull/134/commits/a3996b4ec9f290724394ccaeffe17846c1003c9c).

crates/orchestrator/src/tests/database/mod.rs
heemankv: Valid, added a [panic else case](https://github.com/madara-alliance/madara-orchestrator/pull/134/commits/e678708abeaa7175eed692b07d6c2c02fbe88f40).

crates/orchestrator/src/tests/database/mod.rs
heemankv: Alright, cleaned [here](https://github.com/madara-alliance/madara-orchestrator/pull/134/commits/a9943aeacfce70762692a00182f7f0562a44ff24).

crates/orchestrator/src/jobs/mod.rs
heemankv: [Here](https://github.com/madara-alliance/madara-orchestrator/pull/134/commits/962ce2151296389dcdd85bcc5ccc1535c1e39efa) you go !

crates/orchestrator/src/jobs/types.rs
```diff
@@ -138,6 +140,56 @@ pub struct JobItem {
...
+    pub internal_id: Option<String>,
+    pub job_type: Option<JobType>,
+    pub status: Option<JobStatus>,
+    pub external_id: Option<ExternalId>,
+    pub metadata: Option<HashMap<String, String>>,
+}
+
+/// implements default selections for JobItemUpdates
+impl Default for JobItemUpdates {
+    fn default() -> Self {
```
apoorvsadana, line 157: instead of default, let's rename this to `new`

crates/orchestrator/src/jobs/types.rs
apoorvsadana: no we can skip that

crates/orchestrator/src/jobs/types.rs
apoorvsadana: i mean what if it's not a document? we still update version and updated_at?

heemankv: Valid, resolved [here](https://github.com/madara-alliance/madara-orchestrator/pull/134/commits/881b3aac10f025ffe5e81fa15f80703e57c667dc).

crates/orchestrator/src/jobs/types.rs
heemankv: Alright, changes [here](https://github.com/madara-alliance/madara-orchestrator/pull/134/commits/881b3aac10f025ffe5e81fa15f80703e57c667dc).

crates/orchestrator/src/database/mongodb/mod.rs
```diff
@@ -68,11 +68,13 @@ impl MongoDb {
             if !is_update_available {
                 return Err(("No field to be updated, likely a false call"));
             }
+            
+            // Add additional fields that are always updated
+            doc.insert("version", Bson::Int32(current_job.version + 1));
```
apoorvsadana, line 68: do we not throw an error if it's not a document?

apoorvsadana (approved):

ocdbytes (approved):
lgtm

PR #145, opened 2024-10-09
refactor env

apoorvsadana:


.env.test
```diff
@@ -64,14 +62,17 @@ L1_CORE_CONTRACT_ADDRESS="0xE2Bb56ee936fd6433DC0F6e7e3b8365C906AA057"
...
 ## Instrumentation
 OTEL_SERVICE_NAME="madara_orchestrator"
 
 ##### Tests #####
 
 STARKNET_OPERATOR_ADDRESS="0x5b98B836969A60FEC50Fa925905Dd1D382a7db43"
+AWS_SNS_ARN_NAME="madara-orchestrator-arn"
+MADARA_BINARY_PATH="/path/to/madara"
+# pick up by AWS sdk
+AWS_ENDPOINT_URL="http://localhost.localstack.cloud:4566"
```
heemankv, line 78: ENV : 
- `AWS_ENDPOINT_URL` 
- `MADARA_BINARY_PATH`
- `AWS_SNS_ARN_NAME`
- `STARKNET_OPERATOR_ADDRESS`
are a part of `.env.test` but not found in `.env.example`

crates/da-clients/ethereum/src/config.rs
```diff
@@ -12,17 +12,11 @@ use crate::EthereumDaClient;
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct EthereumDaConfig {
     pub rpc_url: String,
-    pub memory_pages_contract: String,
```
heemankv, line 15: Context on why these were used / are removed now would be appreciated.

.env.test
apoorvsadana: only used for tests, it's under the tests section

crates/da-clients/ethereum/src/config.rs
apoorvsadana: it's not needed. i guess it was added in the very beginning when it was copied from Madara but never removed it

heemankv: Alright.

heemankv (approved):

Mohiiit (approved):

PR #149, opened 2024-10-10
Mongo db migrations js

apoorvsadana:
Not the best code because

1. It hardcodes mongodb in the Dockferfile
2. It needs node/npm to be installed in Docker just to run a migration

There was one [Rust library](https://docs.rs/mongodb-migrator/latest/mongodb_migrator/) that handles mongodb migrations but it doesn't look maintained. Ideally, we can move migrations to a generic function that databases need to implement and we can call them in `init_config`. Or we can just do the migrations in `new_with_settings`. That would however, need custom logic to be added so I did this for now as it does the job.


Dockerfile
```diff
@@ -54,12 +56,15 @@ WORKDIR /usr/src/madara-orchestrator
...
+# Install Node.js dependencies for migrations
+RUN npm install
+
 
 FROM debian:bookworm
 
 # Install runtime dependencies
 RUN apt-get -y update && \
-    apt-get install -y openssl ca-certificates &&\
+    apt-get install -y openssl ca-certificates nodejs npm &&\
```
Mohiiit, line 67: already installed above right?
```suggestion
    apt-get install -y openssl ca-certificates &&\
```

package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "madara-orchestrator",
+  "version": "1.0.0",
+  "description": "The Madara orchestrator is designed to be an additional service which runs in parallel to Madara and handles",
```
Mohiiit, line 4: nitpick: handles .... ?

Dockerfile
apoorvsadana: that was in the builder stage, you need to install it again here

Mohiiit (approved):

heemankv (commented):
LGTM!

heemankv (approved):

PR #152, opened 2024-10-10
atomic create job

apoorvsadana:


crates/orchestrator/src/database/mongodb/mod.rs
```diff
@@ -87,9 +56,37 @@ impl MongoDb {
 #[async_trait]
 impl Database for MongoDb {
     #[tracing::instrument(skip(self), fields(function_type = "db_call"))]
-    async fn create_job(&self, job: JobItem) -> Result<JobItem> {
-        self.get_job_collection().insert_one(&job, None).await?;
-        Ok(job)
+    async fn create_job(&self, job: JobItem) -> Result<JobItem, JobError> {
+        let options = UpdateOptions::builder().upsert(true).build();
+
+        let updates = job.to_document().map_err(|e| JobError::Other(e.into()))?;
```
heemankv, line 62: IMO, can we not read the job_type and internal_id from job itself, is it necessary to get it from updates ?
I would prefer to do less operations on Document and more on typed Structs.
WDYT?

migrations/00000000000000-init.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
       .collection("jobs")
       .createIndexes([
         { key: { id: 1 } },
-        { key: { job_type: 1, internal_id: -1 } },
+        { key: { job_type: 1, internal_id: -1 }, unique: true },
         { key: { job_type: 1, status: 1, internal_id: -1 } },
```
heemankv, line 9: Does it make sense to not have `unique` on `{ key: { job_type: 1, status: 1, internal_id: -1 } },`
Could you give an example?

ocdbytes (approved):
lgtm

crates/orchestrator/src/database/mongodb/mod.rs
apoorvsadana: I tried that first but I had to convert that job type to a bson as well to use inside `doc!`, so since the conversion was already happening above I can use it directly.

migrations/00000000000000-init.js
apoorvsadana: I don't think so. If `job_type` and `internal_id` are already unique then `job_type`, `internal_id` and `status` will also always be unique. I don't see an advantaged in asking MongoDB to check for that as well (or maybe MongoDB optimises and skips it itself). Am I missing something?

PR #158, opened 2024-10-14
Add multiple queue based on job type

byteZorvin:
- This add multiple queues for processing and verification based on the job type.
- This is done because after a particular job was completed say snos_run a new job is created like Proving job which goes to end to of the queue which will process only after all the current jobs would have been processed(say there are 10 snos run job already in the queue).
- After creating multiple queues based on job type we can more evenly distribute compute resources across each jobs type


.env.test
```diff
@@ -21,8 +21,22 @@ AWS_S3_BUCKET_NAME="madara-orchestrator-test-bucket"
...
-SQS_JOB_VERIFICATION_QUEUE_URL="http://sqs.us-east-1.localhost.localstack.cloud:4566/000000000000/madara_orchestrator_job_verification_queue"
+
+SQS_SNOS_JOB_PROCESSING_QUEUE_URL="http://sqs.us-east-1.localhost.localstack.cloud:4566/000000000000/madara_orchestrator_snos_job_processing_queue"
+SQS_SNOS_JOB_VERIFICATION_QUEUE_URL="http://sqs.us-east-1.localhost.localstack.cloud:4566/000000000000/madara_orchestrator_snos_job_verification_queue"
+
+SQS_PROVING_JOB_PROCESSING_QUEUE_URL="http://sqs.us-east-1.localhost.localstack.cloud:4566/000000000000/madara_orchestrator_proving_job_processing_queue"
+SQS_PROVING_JOB_VERIFICATION_QUEUE_URL="http://sqs.us-east-1.localhost.localstack.cloud:4566/000000000000/madara_orchestrator_proving_job_verification_queue"
+
+SQS_PROOF_REGISTRATION_JOB_PROCESSING_QUEUE_URL="http://sqs.us-east-1.localhost.localstack.cloud:4566/000000000000/madara_orchestrator_proof_registration_job_processing_queue"
+SQS_PROOF_REGISTRATION_JOB_VERIFICATION_QUEUE_URL="http://sqs.us-east-1.localhost.localstack.cloud:4566/000000000000/madara_orchestrator_proof_registration_job_verification_queue"
```
apoorvsadana, line 32: can delete this for now

apoorvsadana (approved):

crates/orchestrator/src/queue/job_queue.rs
```diff
@@ -20,9 +21,23 @@ use crate::workers::snos::SnosWorker;
...
+pub const PROOF_REGISTRATION_JOB_VERIFICATION_QUEUE: &str =
+    "madara_orchestrator_proof_registration_job_verification_queue";
+
+pub const DATA_SUBMISSION_JOB_PROCESSING_QUEUE: &str = "madara_orchestrator_data_submission_job_processing_queue";
+pub const DATA_SUBMISSION_JOB_VERIFICATION_QUEUE: &str = "madara_orchestrator_data_submission_job_verification_queue";
+
+pub const UPDATE_STATE_JOB_PROCESSING_QUEUE: &str = "madara_orchestrator_update_state_job_processing_queue";
+pub const UPDATE_STATE_JOB_VERIFICATION_QUEUE: &str = "madara_orchestrator_update_state_job_verification_queue";
+
+// Below is the Dead Letter Queue for the above two jobs.
```
heemankv, line 40: nitpick: change to
Below is the single Dead Letter Queue for all the above queues.

crates/orchestrator/src/queue/job_queue.rs
```diff
@@ -102,14 +117,47 @@ enum DeliveryReturnType {
     NoMessage,
 }
 
-pub async fn add_job_to_process_queue(id: Uuid, config: Arc<Config>) -> EyreResult<()> {
+pub async fn add_job_to_process_queue(id: Uuid, job_type: &JobType, config: Arc<Config>) -> EyreResult<()> {
     tracing::info!("Adding job with id {:?} to processing queue", id);
-    add_job_to_queue(id, JOB_PROCESSING_QUEUE.to_string(), None, config).await
+
+    match job_type {
```
heemankv, line 123: I think we can simplify this by creating a trait here that matches jobType to it's process/verify queue name.

```RUST 
trait QueueName {
    fn process_queue_name(&self) -> String;
    fn verify_queue_name(&self) -> String;
}

impl QueueName for JobType {
    fn process_queue_name(&self) -> String {
        match self {
            JobType::SnosRun => SNOS_JOB_PROCESSING_QUEUE,
            JobType::ProofCreation => PROVING_JOB_PROCESSING_QUEUE,
            // so on . . .           
        }.to_string()
    }
    fn verify_queue_name(&self)- > String {
        // similar
    }
}
```

Then you can just do :
`add_job_to_queue(id, job_type.process_queue_name(), None, config).await;`
inside 
`add_job_to_process_queue` or `add_job_to_verify_queue`


This looks cleaner to me 
WDYT @byteZorvin ?

crates/orchestrator/src/queue/job_queue.rs
```diff
@@ -102,14 +117,47 @@ enum DeliveryReturnType {
...
+pub async fn add_job_to_verification_queue(
+    id: Uuid,
+    job_type: &JobType,
+    delay: Duration,
+    config: Arc<Config>,
+) -> EyreResult<()> {
     tracing::info!("Adding job with id {:?} to verification queue", id);
-    add_job_to_queue(id, JOB_VERIFICATION_QUEUE.to_string(), Some(delay), config).await
+
+    match job_type {
```
heemankv, line 146: Same as above, implementing QueueName

crates/orchestrator/src/queue/job_queue.rs
```diff
@@ -322,9 +370,48 @@ macro_rules! spawn_consumer {
...
+        config.clone()
+    );
+    spawn_consumer!(
+        DATA_SUBMISSION_JOB_VERIFICATION_QUEUE.to_string(),
+        verify_job,
+        consume_job_from_queue,
+        config.clone()
+    );
+
+    spawn_consumer!(
```
heemankv, line 392: I am not sure but are we using proof_registration queues anywhere ? 
If not let's mark a todo instead of spawning them ? 
wdyt @byteZorvin

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -101,7 +104,7 @@ async fn create_job_job_exists_in_db_works() {
 
     // Queue checks.
     let consumed_messages =
-        services.config.queue().consume_message_from_queue(JOB_PROCESSING_QUEUE.to_string()).await.unwrap_err();
+        services.config.queue().consume_message_from_queue(PROVING_JOB_PROCESSING_QUEUE.to_string()).await.unwrap_err();
     assert_matches!(consumed_messages, QueueError::NoData);
```
heemankv, line 108: The trait QueueName makes sense here since now the code is highly dependent on the developer
We could create a variable with job type:
```RUST
let job_type = JobType::ProofCreation;
build_job_item_by_type_and_status(job_type, JobStatus::Created, "0".to_string());
// rest code
services.config.queue().consume_message_from_queue(job_type.processing_queue_name()).await.unwrap_err();
```
This is much less depended on the developer now and creates a single point of failure.

This needs to be changed in all the test cases you are using any type of queue.

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -72,7 +75,7 @@ async fn create_job_job_does_not_exists_in_db_works() {
 
     // Queue checks.
     let consumed_messages =
-        services.config.queue().consume_message_from_queue(JOB_PROCESSING_QUEUE.to_string()).await.unwrap();
+        services.config.queue().consume_message_from_queue(SNOS_JOB_PROCESSING_QUEUE.to_string()).await.unwrap();
```
heemankv, line 78: Queue_Name can be used here.

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -130,19 +133,20 @@ async fn create_job_job_handler_is_not_implemented_panics() {
 
     // Queue checks.
     let consumed_messages =
-        services.config.queue().consume_message_from_queue(JOB_PROCESSING_QUEUE.to_string()).await.unwrap_err();
+        services.config.queue().consume_message_from_queue(PROVING_JOB_PROCESSING_QUEUE.to_string()).await.unwrap_err();
```
heemankv, line 136: Queue_Name can be used here.

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -130,19 +133,20 @@ async fn create_job_job_handler_is_not_implemented_panics() {
...
     assert_matches!(consumed_messages, QueueError::NoData);
 }
 
 /// Tests `process_job` function when job is already existing in the db and job status is either
 /// `Created` or `VerificationFailed`.
 #[rstest]
-#[case(JobType::SnosRun, JobStatus::Created)]
-#[case(JobType::DataSubmission, JobStatus::VerificationFailed)]
+#[case(JobType::SnosRun, JobStatus::Created, SNOS_JOB_VERIFICATION_QUEUE)]
+#[case(JobType::DataSubmission, JobStatus::VerificationFailed, DATA_SUBMISSION_JOB_VERIFICATION_QUEUE)]
```
heemankv, line 144: nitpick: I don't think we can use the Queue_Name here, 
wdyt @byteZorvin ?

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -180,7 +184,7 @@ async fn process_job_with_job_exists_in_db_and_valid_job_processing_status_works
 
     // Queue checks
     let consumed_messages =
-        services.config.queue().consume_message_from_queue(JOB_VERIFICATION_QUEUE.to_string()).await.unwrap();
+        services.config.queue().consume_message_from_queue(verification_queue.to_string()).await.unwrap();
```
heemankv, line 187: Queue_Name can be used here.

heemankv (changes requested):

.env.test
byteZorvin: Deleted

crates/orchestrator/src/queue/job_queue.rs
byteZorvin: updated

crates/orchestrator/src/queue/job_queue.rs
byteZorvin: Agreed updated !!

crates/orchestrator/src/tests/jobs/mod.rs
byteZorvin: updated

crates/orchestrator/src/tests/jobs/mod.rs
byteZorvin: updated

crates/orchestrator/src/tests/jobs/mod.rs
byteZorvin: Would want to to keep the case with specific values, personal preference though

crates/orchestrator/src/tests/jobs/mod.rs
byteZorvin: It cant be used here, its a case value from the args.

crates/orchestrator/src/queue/job_queue.rs
byteZorvin: Removed it

crates/orchestrator/src/tests/jobs/mod.rs
byteZorvin: My bad we can use it, updated

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -477,8 +491,12 @@ async fn verify_job_with_rejected_status_works() {
     sleep(Duration::from_secs(5)).await;
 
     // Queue checks.
-    let consumed_messages_processing_queue =
-        services.config.queue().consume_message_from_queue(JOB_PROCESSING_QUEUE.to_string()).await.unwrap_err();
+    let consumed_messages_processing_queue = services
+        .config
+        .queue()
+        .consume_message_from_queue(DATA_SUBMISSION_JOB_PROCESSING_QUEUE.to_string())
```
heemankv, line 497: Any reason as to why we have not used job_item.job_type.process_queue_name() function here and similar places ? What am I missing ?
@byteZorvin

heemankv (approved):

PR #159, opened 2024-10-15
move failed jobs to the failed state

apoorvsadana:
Failed jobs were getting left in `LockedForProcessing` state. This is incorrect, every job should move to a terminal state after everything is done. In this PR, the following changes are done

1. If processing fails, move job to the failed state
2. If verification fails and no more process attempts are left, move job to the failed state

A failure reason metadata field is added for debugging eventually.

The PR also does an optimisation in update_job which now returns the updated item (vs before where we manually updated the job in the calling function which was hacky)

This PR does NOT fix the issue with jobs being left in `LockedForProcessing` when an orchestrator pod gets randomly killed. Still need to figure out what to do here.

### DIFF
--- a/crates/orchestrator/src/database/mongodb/mod.rs
+++ b/crates/orchestrator/src/database/mongodb/mod.rs
@@ -145,7 +145,10 @@ impl Database for MongoDb {
 
         let result = self.get_job_collection().find_one_and_update(filter, update, options).await?;
         match result {
-            Some(job) => Ok(job),
+            Some(job) => {
+                tracing::debug!(job_id = %current_job.id, category = "db_call", "Job updated successfully");
+                Ok(job)
+            },
             None => {
                 tracing::warn!(job_id = %current_job.id, category = "db_call", "Failed to update job. Job version is likely outdated");
                 return Err(eyre!("Failed to update job. Job version is likely outdated"));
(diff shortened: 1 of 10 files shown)


ocdbytes (approved):
lgtm !!

crates/orchestrator/src/database/mongodb/mod.rs
```diff
@@ -140,14 +143,14 @@ impl Database for MongoDb {
             "$set": non_null_updates
         };
 
-        let result = self.get_job_collection().update_one(filter, update, options).await?;
-        if result.modified_count == 0 {
-            tracing::warn!(job_id = %current_job.id, category = "db_call", "Failed to update job. Job version is likely outdated");
-            return Err(eyre!("Failed to update job. Job version is likely outdated"));
+        let result = self.get_job_collection().find_one_and_update(filter, update, options).await?;
+        match result {
+            Some(job) => Ok(job),
```
Mohiiit, line 148: tracing debug that the job has been updated successfully?

crates/orchestrator/src/jobs/mod.rs
```diff
@@ -220,19 +221,24 @@ pub async fn process_job(id: Uuid, config: Arc<Config>) -> Result<(), JobError>
 
     tracing::debug!(job_id = ?id, job_type = ?job.job_type, "Getting job handler");
     let job_handler = factory::get_job_handler(&job.job_type).await;
-    let external_id = job_handler.process_job(config.clone(), &mut job).await?;
+    let external_id = match job_handler.process_job(config.clone(), &mut job).await {
+        Ok(external_id) => external_id,
+        Err(e) => {
```
Mohiiit, line 226: wdyt having trace debug here might help?

apoorvsadana: added!

Mohiiit (approved):

PR #161, opened 2024-10-16
Change layout of snos

byteZorvin:
- Currently snos is being run in `all_cairo` layout which is not provable. The correct layout of snos should be `stakrnet_with_keccak`.
- This pr fixes the above issue


Mohiiit (approved):

apoorvsadana:
Closing as we realized it needs to be `all_cairo` only

PR #164, opened 2024-10-21
Endpoints to trigger Process and Verify Jobs

heemankv:
This PR adds two endpoints `/trigger/process_job` and `/trigger/verify_job` which takes an id as query param, if id refers to a valid unprocessed/unverified job, it triggers for the same.

### DIFF
--- a/crates/orchestrator/src/routes/job_routes.rs
+++ b/crates/orchestrator/src/routes/job_routes.rs
@@ -1,0 +1,71 @@
+use std::sync::Arc;
+
+use axum::extract::{Query, State};
+use axum::http::StatusCode;
+use axum::response::{IntoResponse, Response};
+use axum::routing::get;
+use axum::Router;
+use serde::Deserialize;
+use uuid::Uuid;
+
+use crate::config::Config;
+use crate::jobs::{process_job, verify_job, JobError};
+
+#[derive(Deserialize)]
+struct JobParams {
+    id: String,
+}
+
+struct JobResult(Result<(), JobError>);
+
+impl IntoResponse for JobResult {
+    fn into_response(self) -> Response {
+        match self.0 {
+            Ok(_) => (StatusCode::OK, "Job processing completed.").into_response(),
+            Err(e) => (StatusCode::INTERNAL_SERVER_ERROR, format!("Job processing failed: {:?}", e)).into_response(),
+        }
+    }
+}
+
+async fn handle_process_job_request(
+    Query(params): Query<JobParams>,
+    State(config): State<Arc<Config>>,
+) -> impl IntoResponse {
+    let id = match Uuid::parse_str(&params.id) {
+        Ok(id) => id,
+        Err(_) => return (StatusCode::BAD_REQUEST, "Invalid UUID").into_response(),
+    };
+
+    JobResult(process_job(id, config).await).into_response()
+}
+
+async fn handle_verify_job_request(
+    Query(params): Query<JobParams>,
+    State(config): State<Arc<Config>>,
+) -> impl IntoResponse {
+    let id = match Uuid::parse_str(&params.id) {
+        Ok(id) => id,
+        Err(_) => return (StatusCode::BAD_REQUEST, "Invalid UUID").into_response(),
+    };
+
+    JobResult(verify_job(id, config).await).into_response()
+}
+
+pub fn job_routes(config: Arc<Config>) -> Router {
+    Router::new()
+        .route(
+            "/process-job",
+            get({
+                let shared_state = config.clone();
+                move |path| handle_process_job_request(path, State(shared_state))
+            }),
+        )
+        .route(
+            "/verify-job",
+            get({
+                let shared_state = config.clone();
+                move |path| handle_verify_job_request(path, State(shared_state))
+            }),
+        )
+        .with_state(config)
+}
--- a/crates/orchestrator/src/routes/job_routes.rs
+++ b/crates/orchestrator/src/routes/job_routes.rs
@@ -53,19 +53,23 @@ async fn handle_verify_job_request(
 
 pub fn job_routes(config: Arc<Config>) -> Router {
     Router::new()
-        .route(
-            "/process-job",
-            get({
-                let shared_state = config.clone();
-                move |path| handle_process_job_request(path, State(shared_state))
-            }),
-        )
-        .route(
-            "/verify-job",
-            get({
-                let shared_state = config.clone();
-                move |path| handle_verify_job_request(path, State(shared_state))
-            }),
+        .nest(
+            "/trigger",
+            Router::new()
+                .route(
+                    "/process-job",
+                    get({
+                        let shared_state = config.clone();
+                        move |path| handle_process_job_request(path, State(shared_state))
+                    }),
+                )
+                .route(
+                    "/verify-job",
+                    get({
+                        let shared_state = config.clone();
+                        move |path| handle_verify_job_request(path, State(shared_state))
+                    }),
+                ),
         )
         .with_state(config)
 }
--- a/crates/orchestrator/src/tests/server/job_routes.rs
+++ b/crates/orchestrator/src/tests/server/job_routes.rs
@@ -1,8 +1,9 @@
-pub mod job_routes;
 use std::io::Read;
 use std::net::SocketAddr;
+use std::sync::Arc;
 
 use axum::http::StatusCode;
+use chrono::{SubsecRound as _, Utc};
 use hyper::body::Buf;
 use hyper::{Body, Request};
 use rstest::*;
@@ -10,13 +11,16 @@ use starknet::providers::jsonrpc::HttpTransport;
 use starknet::providers::JsonRpcClient;
 use url::Url;
 use utils::env_utils::get_env_var_or_default;
+use uuid::Uuid;
 
+use crate::config::{init_config, Config};
+use crate::jobs::types::{ExternalId, JobItem, JobStatus, JobType};
 use crate::queue::init_consumers;
-use crate::routes::routes::app_router;
+use crate::routes::job_routes::job_routes;
 use crate::tests::config::TestConfigBuilder;
 
 #[fixture]
-pub async fn setup_app_server() -> SocketAddr {
+pub async fn setup_job_server() -> (SocketAddr, Arc<Config>) {
     let provider = JsonRpcClient::new(HttpTransport::new(
         Url::parse("http://localhost:9944".to_string().as_str()).expect("Failed to parse URL"),
     ));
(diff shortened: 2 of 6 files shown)


crates/orchestrator/src/tests/server/job_routes.rs
```diff
@@ -0,0 +1,102 @@
...
+use crate::config::{init_config, Config};
+use crate::jobs::types::{ExternalId, JobItem, JobStatus, JobType};
+use crate::queue::init_consumers;
+use crate::routes::job_routes::job_routes;
+use crate::tests::config::TestConfigBuilder;
+
+#[fixture]
+pub async fn setup_job_server() -> (SocketAddr, Arc<Config>) {
+    let provider = JsonRpcClient::new(HttpTransport::new(
+        Url::parse("http://localhost:9944".to_string().as_str()).expect("Failed to parse URL"),
```
Mohiiit, line 25: let's move this url to env?

crates/orchestrator/src/routes/job_routes.rs
```diff
@@ -0,0 +1,75 @@
...
+async fn handle_process_job_request(
+    Query(params): Query<JobParams>,
+    State(config): State<Arc<Config>>,
+) -> impl IntoResponse {
+    let id = match Uuid::parse_str(&params.id) {
+        Ok(id) => id,
+        Err(_) => return (StatusCode::BAD_REQUEST, "Invalid UUID").into_response(),
+    };
+
+    JobResult(process_job(id, config).await).into_response()
```
Mohiiit, line 39: so if I am not wrong, this will finally either returns a ok or internal server error right?

Mohiiit, line 39: so if I am not wrong, this will finally either returns a ok or internal server error right?

crates/orchestrator/src/routes/job_routes.rs
```diff
@@ -0,0 +1,75 @@
...
+struct JobResult(Result<(), JobError>);
+
+impl IntoResponse for JobResult {
+    fn into_response(self) -> Response {
+        match self.0 {
+            Ok(_) => (StatusCode::OK, "Job processing completed.").into_response(),
+            Err(e) => (StatusCode::INTERNAL_SERVER_ERROR, format!("Job processing failed: {:?}", e)).into_response(),
+        }
+    }
+}
```
Mohiiit, line 28: Currently the code throws either okay or internal error, which might be suitable for the needs, this is my suggestion in case we want to have a very descriptive error handling, ofc if you implement this, you have to change a bunch of other stuff as well (mostly the input and the output ofc)

so lmk wdyt?
```suggestion
#[derive(Debug, Serialize)]
struct ApiResponse<T> {
    data: Option<T>,
    error: Option<String>,
}

enum ApiError {
    InvalidUuid(String),
    JobProcessingFailed(JobError),
}

impl IntoResponse for ApiError {
    fn into_response(self) -> Response {
        let (status, error_message) = match self {
            Self::InvalidUuid(id) => (StatusCode::BAD_REQUEST, format!("Invalid UUID: {}", id)),
            Self::JobProcessingFailed(e) => (StatusCode::INTERNAL_SERVER_ERROR, format!("Job processing failed: {:?}", e)),
        };
        
        let body = Json(ApiResponse {
            data: None,
            error: Some(error_message),
        });
        
        (status, body).into_response()
    }
}
```
```suggestion
struct JobResult(Result<(), JobError>);

impl IntoResponse for JobResult {
    fn into_response(self) -> Response {
        match self.0 {
            Ok(_) => (StatusCode::OK, "Job processing completed.").into_response(),
            Err(e) => (StatusCode::INTERNAL_SERVER_ERROR, format!("Job processing failed: {:?}", e)).into_response(),
        }
    }
}
```

crates/orchestrator/src/routes/job_routes.rs
```diff
@@ -0,0 +1,75 @@
...
+
+async fn handle_process_job_request(
+    Query(params): Query<JobParams>,
+    State(config): State<Arc<Config>>,
+) -> impl IntoResponse {
```
Mohiiit, line 33:
```suggestion
) -> Result<impl IntoResponse, ApiError> {
```

crates/orchestrator/src/routes/job_routes.rs
```diff
@@ -0,0 +1,75 @@
...
+                .route(
+                    "/verify-job",
+                    get({
+                        let shared_state = config.clone();
+                        move |path| handle_verify_job_request(path, State(shared_state))
+                    }),
+                ),
+        )
+        .with_state(config)
+}
```
Mohiiit, line 75: Something like this looks cleaner right, wdyt? maybre a lil tricky to read tho
```suggestion
pub fn job_routes(config: Arc<Config>) -> Router {
    Router::new()
        .nest("/trigger", trigger_routes(config.clone()))
        .with_state(config)
}

fn trigger_routes(config: Arc<Config>) -> Router {
    Router::new()
        .route("/process-job", get(handle_process_job_request))
        .route("/verify-job", get(handle_verify_job_request))
        .with_state(config)
}
```

crates/orchestrator/src/tests/server/job_routes.rs
```diff
@@ -0,0 +1,102 @@
...
+    ));
+
+    TestConfigBuilder::new().configure_starknet_client(provider.into()).build().await;
+
+    let host = get_env_var_or_default("HOST", "127.0.0.1");
+    let port = get_env_var_or_default("PORT", "3000").parse::<u16>().expect("PORT must be a u16");
+    let address = format!("{}:{}", host, port);
+
+    let listener = tokio::net::TcpListener::bind(address.clone()).await.expect("Failed to get listener");
+    let addr = listener.local_addr().unwrap();
```
Mohiiit, line 35: remove unwrap

Mohiiit (changes requested):

crates/orchestrator/src/routes/job_routes.rs
heemankv: Yes since the `JobResult` will be given to `IntoResponse` which changes the `Result` to a Response status code of 200 or 500

heemankv: Yes, same as above.

crates/orchestrator/src/routes/job_routes.rs
heemankv: I think your approach is cleaner, thanks for the suggestion.

crates/orchestrator/src/tests/server/job_routes.rs
heemankv: I personally think `unwraps` in `tests` are not a problem, currently changing it to `expect` for a better error message, lemme know what you think @Mohiiit

crates/orchestrator/src/tests/server/job_routes.rs
heemankv: Valid, changed [here](https://github.com/madara-alliance/madara-orchestrator/pull/164/commits/ec1f2470778c1e75b478f2411eba08bfba300075).

crates/orchestrator/src/tests/server/job_routes.rs
Mohiiit: yes true that I think, you can ignore for these changes

crates/orchestrator/src/routes/job_routes.rs
Mohiiit: ideally we should have bad request or internal server error I think, just to be a clear

Mohiiit (approved):

crates/orchestrator/src/routes/job_routes.rs
```diff
@@ -0,0 +1,68 @@
...
+
+use crate::config::Config;
+use crate::jobs::{process_job, verify_job, JobError};
+
+#[derive(Deserialize)]
+struct JobParams {
+    id: String,
+}
+
+struct JobResult(Result<(), JobError>);
```
apoorvsadana, line 19: why not have a struct that is returned to the user here? instead of passing a string `Job processing completed.`

crates/orchestrator/src/routes/job_routes.rs
```diff
@@ -0,0 +1,68 @@
...
+    JobResult(verify_job(id, config).await).into_response()
+}
+
+pub fn job_routes(config: Arc<Config>) -> Router {
+    Router::new().nest("/trigger", trigger_routes(config.clone()))
+}
+
+fn trigger_routes(config: Arc<Config>) -> Router {
+    Router::new()
+        .route("/process-job", get(handle_process_job_request))
```
apoorvsadana, line 65: our endpoints should follow convention. sharing quick response from claude

```
For job-related operations like processing and verification, here are the conventional endpoint designs:

Process Job:

POST /jobs/{jobId}/process

Verify Job:

POST /jobs/{jobId}/verify
This follows RESTful conventions because:

We use the noun jobs as the resource
The job ID is part of the path parameter
The actions (process/verify) are verbs that come after the resource identifier
We use POST since these are actions that modify the state of the job
```

crates/orchestrator/src/tests/server/job_routes.rs
```diff
@@ -0,0 +1,168 @@
...
+use crate::jobs::job_handler_factory::mock_factory;
+use crate::jobs::types::{ExternalId, JobItem, JobStatus, JobType, JobVerificationStatus};
+use crate::jobs::{Job, MockJob};
+use crate::queue::init_consumers;
+use crate::routes::app_routes::{app_router, handler_404};
+use crate::routes::job_routes::job_routes;
+use crate::tests::config::{ConfigType, TestConfigBuilder};
+
+#[fixture]
+async fn setup_job_server() -> (SocketAddr, Arc<Config>) {
```
apoorvsadana, line 26: can this be shared with the health route test case?

heemankv: Valid, implemented [here](https://github.com/madara-alliance/madara-orchestrator/pull/164/commits/6cddb0712a31d221286b75c8ee398e0e585fc915).

heemankv: Valid, implemented [here](https://github.com/madara-alliance/madara-orchestrator/pull/164/commits/af460988fe4345796a7ab735017e0fe50ef9ed34).

Thanks @Mohiiit for the structure format : [Here](https://github.com/madara-alliance/madara-orchestrator/pull/164#discussion_r1808360110).

crates/orchestrator/src/tests/server/mod.rs
```diff
@@ -11,40 +13,51 @@ use starknet::providers::JsonRpcClient;
...
+pub async fn setup_server() -> (SocketAddr, Arc<Config>) {
     dotenvy::from_filename("../.env.test").expect("Failed to load the .env.test file");
 
     let madara_url = get_env_var_or_panic("MADARA_RPC_URL");
     let provider = JsonRpcClient::new(HttpTransport::new(
         Url::parse(madara_url.as_str().to_string().as_str()).expect("Failed to parse URL"),
     ));
 
-    TestConfigBuilder::new().configure_starknet_client(provider.into()).build().await;
+    let services = TestConfigBuilder::new()
```
apoorvsadana, line 31: the `setup_server` should take `services` as an argument because the health server doesn't need it for example

apoorvsadana (approved):

crates/orchestrator/src/tests/server/mod.rs
heemankv: resolved with new tokio server spin up

PR #165, opened 2024-10-21
added makefile and fixes for testnet launch

apoorvsadana:


.github/workflows/coverage.yml
```diff
@@ -121,18 +121,15 @@ jobs:
...
-          cd /home/runner/.cargo/git/checkouts
-          # Navigating to snos
-          cd snos-59fe8329bb16fe65
-          # Navigating to the build
-          cd af74c75
-          # Activating the venv
-          source ~/cairo_venv/bin/activate
-          # Building the cairo lang repo requirements
-          ./scripts/setup-tests.sh
+          python3 -m venv orchestrator_venv
```
apoorvsadana, line 124: can use make command? (applies to all coverage files)

.gitignore
```diff
@@ -12,3 +12,11 @@ lcov.info
 **/*-SN.zip
 
 node_modules
+
+# env file created by makefile
+.makefile-env
```
apoorvsadana, line 17: can change to `.makefile.json` when we refactor

.gitmodules
```diff
@@ -0,0 +1,16 @@
...
+	url = https://github.com/eqlabs/pathfinder.git
+    ignore = all
+[submodule "madara"]
+	path = madara
+	url = https://github.com/madara-alliance/madara.git
+    ignore = all
+[submodule "madara-bootstrapper"]
+	path = madara-bootstrapper
+	url = https://github.com/madara-alliance/madara-bootstrapper.git
+    ignore = all
```
apoorvsadana, line 16: can you fix linting here

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## Added
 
+- makefile
```
apoorvsadana, line 9: can you elaborate (many of these are my bugs btw)

Cargo.toml
```diff
@@ -108,13 +109,8 @@ cairo-vm = { git = "https://github.com/Moonsong-Labs/cairo-vm", branch = "notles
...
 
 # Snos & Sharp (Starkware)
-# TODO: Update branch to main once the PR is merged (PR #368 in the snos repository)
-starknet-os = { git = "https://github.com/keep-starknet-strange/snos", rev = "af74c7599231c25ad9c342f6888668680a9782d9" }
-prove_block = { git = "https://github.com/keep-starknet-strange/snos", rev = "af74c7599231c25ad9c342f6888668680a9782d9" }
-
-# Madara prover API
-madara-prover-common = { git = "https://github.com/Moonsong-Labs/madara-prover-api", branch = "od/use-latest-cairo-vm" }
-madara-prover-rpc-client = { git = "https://github.com/Moonsong-Labs/madara-prover-api", branch = "od/use-latest-cairo-vm" }
+starknet-os = { git = "https://github.com/keep-starknet-strange/snos", branch = "main" }
```
apoorvsadana, line 112: move to specific commit

Dockerfile
```diff
@@ -82,8 +82,8 @@ COPY --from=builder /usr/src/madara-orchestrator/migrate-mongo-config.js .
 COPY --from=builder /usr/src/madara-orchestrator/migrations ./migrations
 
 # To be fixed by this https://github.com/keep-starknet-strange/snos/issues/404
-RUN mkdir -p /usr/local/cargo/git/checkouts/snos-59fe8329bb16fe65/af74c75/crates/starknet-os/kzg
-COPY ./crates/da-clients/ethereum/trusted_setup.txt /usr/local/cargo/git/checkouts/snos-59fe8329bb16fe65/af74c75/crates/starknet-os/kzg/trusted_setup.txt
+RUN mkdir -p /usr/local/cargo/git/checkouts/snos-59fe8329bb16fe65/662d170/crates/starknet-os/kzg
```
apoorvsadana, line 85: can fix snos building in Docker file above

cairo-lang
apoorvsadana, line 1: how do we avoid these changes? submodules get annoying sometimes because of this stuff, like how i make submodule changes not show in git. i though `ignore` would do it but doesn't look like

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -198,11 +217,20 @@ impl SettlementClient for EthereumSettlementClient {
 
         let max_fee_per_blob_gas: u128 = self.provider.get_blob_base_fee().await?.to_string().parse()?;
 
+        // calculating y_0 point
+        let y_0_point_u256 = convert_stark_bigint_to_u256(
+            bytes_to_u128(&program_output[Y_LOW_POINT_OFFSET]),
+            bytes_to_u128(&program_output[Y_HIGH_POINT_OFFSET]),
+        );
+        let y_0_point_bytes_32 = Bytes32::from(y_0_point_u256.to_be_bytes());
```
apoorvsadana, line 225: you don't need to append type on var name in general, linters show the type

apoorvsadana: you could just use `y_0` at both places (second delcaration wll override the other, which is nice because first one is only used to create the second one)

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -263,11 +291,30 @@ impl SettlementClient for EthereumSettlementClient {
...
+            100,
+            Duration::from_secs(5),
+            3,
+        )
+        .await?;
 
-        Ok(txn_hash)
+        match res {
+            Some(_) => {
+                println!(">>>> txn hash : {:?} Finalized ✅", pending_transaction.tx_hash().to_string());
```
apoorvsadana, line 305: log

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -263,11 +291,30 @@ impl SettlementClient for EthereumSettlementClient {
...
+                log::error!("Txn hash not finalised");
+            }
+        }
+        Ok(pending_transaction.tx_hash().to_string())
+
+        // Prod feature only (may cause issues while testing with anvil)
+        // let txn_hash = pending_transaction.tx_hash().to_string();
+        // self.wait_for_tx_finality(&txn_hash).await?;
+        //
+        // Ok(txn_hash)
```
apoorvsadana, line 317: let's delete if we don't need this

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -337,6 +384,30 @@ impl SettlementClient for EthereumSettlementClient {
     }
 }
 
+impl EthereumSettlementClient {
+    async fn wait_for_transaction_confirmation(
```
apoorvsadana, line 388: refactor as discussed

crates/settlement-clients/ethereum/src/tests/mod.rs
```diff
@@ -298,8 +299,13 @@ mod settlement_client_tests {
         let blob_data_vec = get_blob_data(fork_block_no);
 
         let x_0_value_bytes32 = Bytes32::from(program_output[10]);
+        let y_0_value_low = Bytes32::from(program_output[14]);
+        let y_0_value_high = Bytes32::from(program_output[15]);
+        let y_0_value_u256 =
+            convert_stark_bigint_to_u256(bytes_to_u128(y_0_value_low.as_ref()), bytes_to_u128(y_0_value_high.as_ref()));
+        let y_0_value_bytes_32 = Bytes32::from(y_0_value_u256.to_be_bytes());
```
apoorvsadana, line 306: same, we don't need to add type to var name

crates/settlement-clients/ethereum/src/types.rs
```diff
@@ -22,3 +23,96 @@ pub type EthHttpProvider = FillProvider<
...
+            bytes[index] = value;
+        }
+        bytes
+    }
+
+    #[rstest]
+    #[case(&[], 0u128, "all zeros")]
+    #[case(&[(31, 1)], 1u128, "last byte is 1")]
+    #[case(&[(31, 255)], 255u128, "last byte is 255")]
+    #[case(&[(30, 1), (31, 255)], 511u128, "last two bytes")]
```
apoorvsadana, line 76: you don't need a test_name. you can do `#[case::all_zeros(...)]`

crates/settlement-clients/ethereum/src/types.rs
```diff
@@ -22,3 +23,96 @@ pub type EthHttpProvider = FillProvider<
...
+        #[case] test_name: &str,
+    ) {
+        let bytes = create_test_bytes(byte_values);
+        let result = bytes_to_u128(&bytes);
+        assert_eq!(result, expected, "Failed case: {}", test_name);
+    }
+
+    #[allow(clippy::needless_range_loop)]
+    #[rstest]
+    fn test_bytes_to_u128_ignores_first_16_bytes() {
```
apoorvsadana, line 89: shouldn't this throw an error?

crates/settlement-clients/ethereum/src/types.rs
```diff
@@ -22,3 +23,96 @@ pub type EthHttpProvider = FillProvider<
...
+        let mut bytes = [255u8; 32];
+        for i in 16..32 {
+            bytes[i] = 0;
+        }
+        let result = bytes_to_u128(&bytes);
+        assert_eq!(result, 0);
+    }
+
+    #[allow(clippy::needless_range_loop)]
```
apoorvsadana, line 98: instead of this clippy, we can use the recommended approach. claude says `bytes[16..32].fill(255);` but ig there's also a range method or something

e2e-tests/tests.rs
```diff
@@ -64,6 +64,7 @@ impl Setup {
         let localstack_instance = LocalStack::new().await;
         localstack_instance.setup_sqs().await.unwrap();
         localstack_instance.delete_event_bridge_rule("worker_trigger_scheduled").await.unwrap();
+        // localstack_instance.setup_event_bridge(WorkerTriggerType::Snos).await.unwrap();
```
apoorvsadana, line 67:
```suggestion
```

crates/settlement-clients/ethereum/src/types.rs
```diff
@@ -22,3 +23,96 @@ pub type EthHttpProvider = FillProvider<
...
+        }
+        let result = bytes_to_u128(&bytes);
+        assert_eq!(result, u128::MAX);
+    }
+
+    #[rstest]
+    #[case(&[(20, 1), (25, 2), (31, 3)], (1u128 << 88) | (2u128 << 48) | 3u128)]
+    #[case(&[(16, 1)], 1u128 << 120)]
+    #[case(&[(31, 1), (30, 1), (29, 1)], 0x010101u128)]
+    fn test_bytes_to_u128_complex_cases(#[case] byte_values: &[(usize, u8)], #[case] expected: u128) {
```
apoorvsadana, line 113: this and simple cases can be the same test case, with different #case

scripts/setup_aws.sh
apoorvsadana, line 1: i created this with claude and it didn't work, in case you haven't changed it let's delete

crates/settlement-clients/ethereum/src/types.rs
ocdbytes: no it only sets the value in bytes to 0.

apoorvsadana: what I mean is you're converting a bytes array greater than u128 into u128. also, now that I think of it, function name should have `bytes_be`/`bytes_le`

Dockerfile
```diff
@@ -27,11 +27,11 @@ RUN wget https://bootstrap.pypa.io/get-pip.py \
...
-RUN python3.9 -m venv /usr/local/cairo_venv
-RUN pip3.9 install ecdsa fastecdsa sympy
-RUN pip3.9 install cairo-lang
-
-RUN python3.9 --version && pip3.9 --version
+#RUN python3.9 -m venv /usr/local/cairo_venv
+#RUN pip3.9 install ecdsa fastecdsa sympy
+#RUN pip3.9 install cairo-lang
+#
+#RUN python3.9 --version && pip3.9 --version
```
apoorvsadana, line 34:
```suggestion
```

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -263,11 +298,17 @@ impl SettlementClient for EthereumSettlementClient {
...
 
-        // Prod feature only (may cause issues while testing with anvil)
-        let txn_hash = pending_transaction.tx_hash().to_string();
-        self.wait_for_tx_finality(&txn_hash).await?;
+        let res = self.wait_for_tx_finality(&pending_transaction.tx_hash().to_string()).await?;
 
-        Ok(txn_hash)
+        match res {
+            Some(_) => {
+                log::info!(">>>> txn hash : {:?} Finalized ✅", pending_transaction.tx_hash().to_string());
```
apoorvsadana, line 305:
```suggestion
                log::info!("Txn hash : {:?} Finalized ✅", pending_transaction.tx_hash().to_string());
```

crates/settlement-clients/ethereum/src/lib.rs
```diff
@@ -40,9 +41,20 @@ use alloy::transports::http::Http;
...
+
 pub const ENV_PRIVATE_KEY: &str = "ETHEREUM_PRIVATE_KEY";
+const X_0_POINT_OFFSET: usize = 10;
+const Y_LOW_POINT_OFFSET: usize = 14;
+const Y_HIGH_POINT_OFFSET: usize = Y_LOW_POINT_OFFSET + 1;
+
+// Ethereum Transaction Finality
+const MAX_ATTEMPTS: usize = 100;
+const REQUIRED_BLOCK_CONFIRMATIONS: u64 = 3;
+const SLEEP_DELAY_SECS: u64 = 5;
```
apoorvsadana, line 57:
```suggestion
const MAX_TX_FINALISATION_ATTEMPTS: usize = 100;
const REQUIRED_BLOCK_CONFIRMATIONS: u64 = 3;
const TX_WAIT_SLEEP_DELAY_SECS: u64 = 5;
```

Makefile
```diff
@@ -0,0 +1,177 @@
...
+MADARA_DATA_PATH := $(shell pwd)/build/madara
+
+# Pathfinder
+PATHFINDER_COMMIT := 138140e5fd967ede92806fd62de47c2e6b65712a
+PATHFINDER_DATA_PATH := $(shell pwd)/build/pathfinder
+PATHFINDER_PATH := $(shell pwd)/pathfinder
+
+# Environment file
+ENV_FILE := $(shell pwd)/.makefile.json
+PID_FILE := $(shell pwd)/.pids.json
```
apoorvsadana, line 26: let's use the same file only? like all make file related configs will be here

apoorvsadana, line 26: add to .gitignore

Makefile
```diff
@@ -0,0 +1,177 @@
...
+	if [ ! -f "$(PID_FILE)" ]; then echo '{}' > $(PID_FILE); fi && \
+	jq --arg key "$(1)" --arg value "$(2)" '. + {($key): ($value)}' $(PID_FILE) > $(PID_FILE).tmp && \
+	mv $(PID_FILE).tmp $(PID_FILE)
+endef
+
+define get_pid
+	jq -r '.$(1) // empty' $(PID_FILE)
+endef
+
+setup:
```
apoorvsadana, line 40: getting this error


```
Starting setup process...
Cleaning up processes...
Cleanup completed.
Cleaning up processes...
Cleanup completed.
jq: error: ey/0 is not defined at <top-level>, line 1:
. + {(ey): (alue)}
jq: error: alue/0 is not defined at <top-level>, line 1:
. + {(ey): (alue)}
jq: 2 compile errors
make: *** [setup] Error 3
anvil --block-time 6
```

ocdbytes: this is a shell error when used with bash it resolves. I have added the changes so that this issue doesn't pop up regardless of the terminal used

Makefile
ocdbytes: this approach is much cleaner.

ocdbytes (approved):

Cargo.toml
```diff
@@ -108,13 +109,8 @@ cairo-vm = { git = "https://github.com/Moonsong-Labs/cairo-vm", branch = "notles
 ] }
 
 # Snos & Sharp (Starkware)
-# TODO: Update branch to main once the PR is merged (PR #368 in the snos repository)
```
heemankv, line 111: I think it's high time that we include a TODO-to-issue workflow, because such changes will not be remembered for long.

Makefile
```diff
@@ -0,0 +1,211 @@
...
+	sh ./scripts/save_json.sh $1 $2
+endef
+
+setup:
+	@echo "Starting setup process..."
+	@make cleanup
+	@trap 'make cleanup' EXIT INT TERM
+
+	# @echo "Starting Anvil..."
+	# @make anvil & $(call save_pid,anvil,$$!)
```
heemankv, line 55: Seems like anvil is being used down below at L109, any reason to have this comment here, if not let's please remove it.

Makefile
```diff
@@ -0,0 +1,211 @@
...
+udc:
+	cd $(BOOTSTRAPPER_PATH) && \
+	git checkout $(BOOTSTRAPPER_COMMIT) && \
+	RUST_LOG=debug cargo run --release -- --mode udc
+
+pathfinder:
+	cd $(PATHFINDER_PATH) && \
+	git checkout $(PATHFINDER_COMMIT) && \
+	rm -rf $(PATHFINDER_DATA_PATH) && \
+	cargo run --release --bin pathfinder -- --ethereum.url wss://eth-sepolia.g.alchemy.com/v2/WIUR5JUZXieEBkze6Xs3IOXWhsS840TX --chain-id MADARA_DEVNET --feeder-gateway-url http://localhost:8080/feeder_gateway --gateway-url  http://localhost:8080/gateway --network custom --storage.state-tries archive --data-directory $(PATHFINDER_DATA_PATH)
```
heemankv, line 170: Seems lie the pvt key is being exposed here : wss://eth-sepolia.g.alchemy.com/v2/WIUR5JUZXieEBkze6Xs3IOXWhsS840TX
Can we deprecate this one and set up a env based pick up for this url ?

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -394,6 +395,30 @@ fn da_word(class_flag: bool, nonce_change: Option<Felt>, num_changes: u64) -> Fe
...
+        state_update.storage_diffs.clone().iter().map(|item| item.address).collect();
+
+    let address_to_insert = find_unique_addresses(addresses_in_nonces, addresses_in_storage_diffs);
+
+    for address in address_to_insert {
+        state_update.storage_diffs.push(ContractStorageDiffItem { address, storage_entries: vec![] })
+    }
+}
+
+fn find_unique_addresses(nonce_addresses: Vec<Felt>, storage_diff_addresses: Vec<Felt>) -> Vec<Felt> {
```
heemankv, line 410: This seems to not be the best optimised code,
This has a time complexity of O(n*m), 
I think we can simply achieve O(n+m) by converting storage_diff_addresses to a hashset.



```rust
fn find_unique_addresses(nonce_addresses: Vec<Felt>, storage_diff_addresses: Vec<Felt>) -> Vec<Felt> {
    let storage_set: HashSet<_> = storage_diff_addresses.into_iter().collect();
   
    nonce_addresses
        .into_iter()
        .filter(|addr| !storage_set.contains(addr))
        .collect()
}
```

crates/settlement-clients/ethereum/src/tests/mod.rs
```diff
@@ -167,7 +168,7 @@ mod settlement_client_tests {
         let nonce = ethereum_settlement_client.get_nonce().await.expect("Unable to fetch nonce");
 
         // keeping 11 elements because the code accesses 10th index as program output
```
heemankv, line 170: Better to update the comment and give an explanation as to why the 16th element is accessed now

scripts/purge_sqs.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+# Set the endpoint URL
+ENDPOINT_URL="http://sqs.us-east-1.localhost.localstack.cloud:4566"
+
+# Define an array of queue names
+QUEUES=(
+    "madara_orchestrator_snos_job_processing_queue"
+    "madara_orchestrator_snos_job_verification_queue"
+    "madara_orchestrator_proving_job_processing_queue"
```
heemankv, line 10: I think there are too many places now where we are managing this list. can we make a single source of truth ?

Cargo.toml
apoorvsadana: is this a common thing? haven't seen it before

apoorvsadana: also, in this specific case, we're deleting a TODO btw, not adding one but your point probably stays. although, i do know there are extensions that people normally use to see the complete list of TODOs in the code

Makefile
apoorvsadana: good catch. so i actually planned to remove from here itself because later, maybe you want to use sepolia to set it up. so if you want anvil you set it up separately. although there are other changes you might need to do as we've never tested makefile with sepolia, but that's the mental model.

crates/settlement-clients/ethereum/src/tests/mod.rs
apoorvsadana: updating the comment, where is 16 accessed?

scripts/purge_sqs.sh
apoorvsadana: Fair enough, creating an issue for this

Makefile
apoorvsadana: good catch, fixed

crates/orchestrator/src/jobs/da_job/mod.rs
apoorvsadana: nice fixed

Cargo.toml
heemankv: Yea there are extensions and I personally use them, but I have seen maintainers become more active about a TODO when it's converted to an issue, maybe because it get's shown up in the issues tab which is usually tracked

here's one to try : https://github.com/marketplace/actions/todo-to-issue

crates/settlement-clients/ethereum/src/tests/mod.rs
heemankv: my bad read the code wrong, actually here we are updating the program_output vec size from 11 to 16, it seems to be because of Y_LOW_POINT_OFFSET being 14, but I am lacking the context here, could you provide a liner or 2 ?

Cargo.toml
heemankv: created issue

crates/settlement-clients/ethereum/src/tests/mod.rs
ocdbytes: We shifted to new contract that has latest format and snos also gives the result in that format so we changed the index in order to make the test pass as Y_LOW_POINT_OFFSET and Y_HIGH_POINT_OFFSET are accessed during state update.

heemankv (approved):

PR #172, opened 2024-11-03
Fix l1 bridge upgrade

apoorvsadana:
Only upgrading L2 ETH bridge and not the L1 fails when you withdraw funds. This PR upgrades the L1 bridge as well.


Mohiiit (approved):

ocdbytes (approved):
lgtm !!

PR #173, opened 2024-11-04
Fix : create job error logs

ocdbytes:


.env.test
```diff
@@ -73,7 +73,7 @@ L1_CORE_CONTRACT_ADDRESS="0xE2Bb56ee936fd6433DC0F6e7e3b8365C906AA057"
 
 ##### SNOS #####
 ## This is needed right now because Madara doesn't support getProof
-RPC_FOR_SNOS=""
+RPC_FOR_SNOS="http://localhost:9944"
```
apoorvsadana, line 76:
```suggestion
RPC_FOR_SNOS="http://localhost:9545"
```

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -98,10 +98,6 @@ async fn create_job_job_exists_in_db_works() {
     let database_client = services.config.database();
     database_client.create_job(job_item.clone()).await.unwrap();
 
-    assert!(
-        create_job(JobType::ProofCreation, "0".to_string(), HashMap::new(), services.config.clone()).await.is_err()
-    );
```
apoorvsadana, line 103: check it returns ok instead and check DB that job isn't there imo

ocdbytes: So basically this should should fail because previously we were throwing an error. but now we just log a warning and safe exit from the create_job function.

ocdbytes: In my opinion the is_ok check is not needed here.

apoorvsadana: we can remove is ok check but i think checking it's not present in the DB makes sense, especially because now the code isn't throwing an error

crates/orchestrator/src/jobs/mod.rs
```diff
@@ -165,7 +165,8 @@ pub async fn create_job(
 
     // this is technically a redundant check, we've another check inside `create_job`
     if existing_job.is_some() {
-        return Err(JobError::JobAlreadyExists { internal_id, job_type });
+        log::warn!("Job already exists for internal_id {internal_id:?} and job_type {job_type:?}. Skipping!");
```
heemankv, line 168: this needs to be a `tracing::warn` and not a `log::warn`

heemankv: And I believe JobError can be serialised, so we can do something like : 
```rust
tracing::warn(JobError::JobAlreadyExists { internal_id, job_type });
```
seems cleaner

crates/orchestrator/src/tests/jobs/mod.rs
ocdbytes: in this test the job is being created above in the first line. After that I can call this function and check in the db that there is only one job. wdyt about this ?

heemankv (commented):
Left some comments

crates/orchestrator/src/tests/jobs/mod.rs
apoorvsadana: ya ig that works but why not just do `get_job_by_internal_id_and_type`? you already have the db client and the fn is already there

apoorvsadana (approved):

Mohiiit (approved):

PR #175, opened 2024-11-04
fix: indexing for get_jobs_without_successor

heemankv:
`get_jobs_without_successor` was clocking at 5 seconds query time,
added indexes on : 
- { "job_type": 1, "status": 1 }
- { "job_type": 1, "internal_id": 1 }
now, it's reduces to `30 ms`

### DIFF
--- a/migrations/00000000000000-init.js
+++ b/migrations/00000000000000-init.js
@@ -8,6 +8,9 @@ module.exports = {
         { key: { job_type: 1, internal_id: -1 }, unique: true },
         { key: { job_type: 1, status: 1, internal_id: -1 } },
         { key: { status: 1 } },
+        // primarily for get_jobs_without_successor
+        { key: { job_type: 1, status: 1 } },
+        { key: { job_type: 1, internal_id: 1 } },
       ]);
   },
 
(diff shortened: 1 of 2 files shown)


migrations/00000000000000-init.js
```diff
@@ -1,14 +1,15 @@
...
-        { key: { status: 1 } },
-      ]);
+    await db.collection("jobs").createIndexes([
+      { key: { id: 1 } },
+      { key: { job_type: 1, internal_id: -1 }, unique: true },
+      { key: { job_type: 1, status: 1, internal_id: -1 } },
+      { key: { status: 1 } },
+      // primarily for get_jobs_without_successor
+      { key: { job_type: 1, status: 1 } },
+      { key: { job_type: 1, internal_id: 1 } },
```
apoorvsadana, line 11: we already have `{ job_type: 1, internal_id: -1 }`, are you sure we need this?

heemankv: Yup, initially I tried with only ` { key: { job_type: 1, status: 1 } }` and the query time actually increased from average 51sec to average 58sec and only after adding `{key: { job_type: 1, internal_id: 1 } }` it reduced to `340ms`

ocdbytes (approved):
lgtm

PR #176, opened 2024-11-04
Fix : snos orch blob data

ocdbytes:


crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -396,25 +396,33 @@ fn da_word(class_flag: bool, nonce_change: Option<Felt>, num_changes: u64) -> Fe
 }
 
 fn refactor_state_update(state_update: &mut StateDiff) {
```
apoorvsadana, line 398: i think it can still be made better per say

1. create `existing_storage_addresses` on the top
2. loop over nonce, if not present in `existing_storage_addresses`, insert in `state_update` and `existing_storage_addresses`
3. same for deployed_contracts

So, 1 hash set and 2 loops. Complexity remains same to be precise but it's cleaner. `required_addresses` feels redundant here.

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -570,6 +590,71 @@ pub mod test {
...
+        let initial_storage = state_diff.storage_diffs.clone();
+
+        // Run the function
+        refactor_state_update(&mut state_diff);
+
+        // Verify results
+        assert!(verify_addresses_have_storage_diffs(&state_diff));
+
+        // Verify no duplicate addresses in storage_diffs
+        let unique_addresses: HashSet<_> = state_diff.storage_diffs.iter().map(|item| &item.address).collect();
```
apoorvsadana, line 636: this is being calculated twice right? once inside `verify_addresses_have_storage_diffs` and once here?

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -570,6 +590,71 @@ pub mod test {
...
+            "Number of unique addresses should match number of storage diffs"
+        );
+        assert_eq!(
+            unique_addresses.len(),
+            expected_storage_count,
+            "Number of storage diffs should match expected count"
+        );
+
+        // Verify storage preservation for existing entries
+        for orig_storage in initial_storage {
```
apoorvsadana, line 649: when we check `verify_addresses_have_storage_diffs`, why can't we check this as well?

pathfinder
apoorvsadana, line 1: why do these things come, any idea?

crates/orchestrator/src/jobs/da_job/mod.rs
ocdbytes: Ig one loop is cleaner. the above method seems same only.

pathfinder
ocdbytes: Claude says to avoid this we can add the submodules to .gitignore

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -396,25 +396,28 @@ fn da_word(class_flag: bool, nonce_change: Option<Felt>, num_changes: u64) -> Fe
...
+        ),
+    )
+    .collect();
+
+    // Add missing storage diffs in one batch
+    state_update.storage_diffs.extend(
+        required_addresses
+            .difference(&existing_storage)
+            .map(|&address| ContractStorageDiffItem { address, storage_entries: Vec::new() }),
+    );
```
apoorvsadana, line 415: ok this is definitely cleaner than before but can you check how many clones this is doing? from my understanding we're copying addresses into `existing_storage` and then into `required_addresses`. Instead, we could 

1. loop over an iterator chain of nonces + deployed contracts
2. if not present in `existing_storage`, insert into `existing_storage` and `state_update.storage_diffs`

This way, I think we will have 
1. A smaller loop (currently we loop on `required_addresses` inside `difference` I beleive and required addresses has storage_diffs, nonce and deployed contracts. now we only loop on nonce and deployed contracts.
2. Less memory, you don't need to have `required_addresses`

ocdbytes: what do you think about this approach ?  

```rust
fn refactor_state_update(state_update: &mut StateDiff) {
    let existing_storage: HashSet<_> = state_update.storage_diffs.iter().map(|item| item.address).collect();

    // Collect new addresses, using a HashSet for deduplication
    let new_addresses: HashSet<_> = Iterator::chain(
        state_update.nonces.iter().map(|item| item.contract_address),
        state_update.deployed_contracts.iter().map(|item| item.address),
    )
    .filter(|address| !existing_storage.contains(address))
    .collect();

    // Add new storage diffs in batch
    state_update.storage_diffs.extend(
        new_addresses.into_iter().map(|address| ContractStorageDiffItem { address, storage_entries: Vec::new() }),
    );
}
```

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -616,4 +673,55 @@ pub mod test {
...
+                .map(|(addr, entries)| ContractStorageDiffItem {
+                    address: addr,
+                    storage_entries: entries.into_iter().map(|(key, value)| StorageEntry { key, value }).collect(),
+                })
+                .collect(),
+            deprecated_declared_classes: vec![],
+            declared_classes: vec![],
+            deployed_contracts: deployed_contracts
+                .into_iter()
+                .map(|(addr, _code)| DeployedContractItem { address: addr, class_hash: Default::default() })
```
Mohiiit, line 695:
```suggestion
                .map(|(addr, _class_hash)| DeployedContractItem { address: addr, class_hash: Default::default() })
```

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -616,4 +673,55 @@ pub mod test {
...
+            {
+                assert_eq!(
+                    orig_storage.storage_entries, current_storage.storage_entries,
+                    "Storage entries changed unexpectedly"
+                );
+            }
+        }
+    }
+
+    fn verify_addresses_have_storage_diffs(state_diff: &StateDiff) -> bool {
```
Mohiiit, line 721: this function is not being called in the test

Mohiiit (changes requested):

crates/orchestrator/src/jobs/da_job/mod.rs
ocdbytes: done

crates/orchestrator/src/jobs/da_job/mod.rs
ocdbytes: resolved

crates/orchestrator/src/jobs/da_job/mod.rs
apoorvsadana: looks good!

crates/orchestrator/src/jobs/da_job/mod.rs
Mohiiit: did you add it? I might be looking at the wrong place

ocdbytes: this is being called here : 

https://github.com/madara-alliance/madara-orchestrator/pull/176/files#diff-01a67dcc79495a881151d118f8080aa28ead03df572a5d8e16a8f6eebc326225R622

Mohiiit (approved):

crates/orchestrator/src/jobs/da_job/mod.rs
```diff
@@ -616,4 +669,56 @@ pub mod test {
...
+                    "Storage entries changed unexpectedly"
+                );
+            }
+        }
+
+        let storage_addresses: HashSet<_> = state_diff.storage_diffs.iter().map(|item| &item.address).collect();
+
+        state_diff.nonces.iter().all(|item| storage_addresses.contains(&item.contract_address))
+            && state_diff.deployed_contracts.iter().all(|item| storage_addresses.contains(&item.address))
+    }
```
apoorvsadana, line 723: why not 

```
 state_diff.nonces.iter().all(|item| storage_addresses.contains(&item.contract_address))
            && state_diff.deployed_contracts.iter().all(|item| storage_addresses.contains(&item.address))
            && initial_storage.iter().all(...)
```

PR #178, opened 2024-11-06
Add panic recovery handling for process_job with test coverage

Mohiiit:
**Description:**

Added graceful panic recovery mechanism to process_job to prevent system crashes and ensure proper error handling. This enhancement improves system stability and maintainability.

Changes:
- Implemented panic recovery using defer/recover pattern in process_job
- The job will move to failed state in the panic case

Test Coverage:
- Added test cases to verify panic scenarios are properly caught and handled
- Verified error is returned instead of propagating the panic


crates/orchestrator/src/jobs/mod.rs
```diff
@@ -227,18 +229,23 @@ pub async fn process_job(id: Uuid, config: Arc<Config>) -> Result<(), JobError>
...
+        Ok(Err(e)) => {
             // TODO: I think most of the times the errors will not be fixed automatically
             // if we just retry. But for some failures like DB issues, it might be possible
             // that retrying will work. So we can add a retry logic here to improve robustness.
             tracing::error!(job_id = ?id, error = ?e, "Failed to process job");
             return move_job_to_failed(&job, config.clone(), format!("Processing failed: {}", e)).await;
         }
+        Err(_) => {
+            tracing::error!(job_id = ?id, "Job handler panicked during processing");
+            return move_job_to_failed(&job, config.clone(), "Job handler panicked during processing".to_string())
```
apoorvsadana, line 246: let's add panic error here

apoorvsadana (approved):

crates/orchestrator/src/jobs/mod.rs
Mohiiit: after catching the panic? we should throw an panic error?

also as per the current logic, if a job x fails, job x+1 won't get process? if that's the case, we should panic

crates/orchestrator/src/tests/jobs/mod.rs
```diff
@@ -192,6 +192,49 @@ async fn process_job_with_job_exists_in_db_and_valid_job_processing_status_works
...
+    let ctx = mock_factory::get_job_handler_context();
+    ctx.expect().times(1).with(eq(JobType::SnosRun)).return_once(move |_| Arc::clone(&job_handler));
+
+    assert!(process_job(job_item.id, services.config.clone()).await.is_ok());
+
+    // DB checks - verify the job was moved to failed state
+    let job_in_db = database_client.get_job_by_id(job_item.id).await.unwrap().unwrap();
+    assert_eq!(job_in_db.status, JobStatus::Failed);
+    assert!(
+        job_in_db.metadata.get(JOB_METADATA_FAILURE_REASON).unwrap().contains("Job handler panicked during processing")
```
apoorvsadana, line 234: we can assert `Simulated panic in process_job` as well here after above change

crates/orchestrator/src/jobs/mod.rs
apoorvsadana: no no, error message that goes inside `move_job_to_failed` should include panic error message.

> also as per the current logic, if a job x fails, job x+1 won't get process? if that's the case, we should panic

I think this will be handle by the worker is enabled check (inside run_worker iirc)

Mohiiit: I see got it

ocdbytes (approved):
lgtm !!

crates/orchestrator/src/jobs/mod.rs
Mohiiit: added

crates/orchestrator/src/tests/jobs/mod.rs
Mohiiit: added

apoorvsadana (approved):

PR #181, opened 2024-11-06
Feat : setup scripts

ocdbytes:


crates/orchestrator/src/data_storage/mod.rs
```diff
@@ -21,7 +21,10 @@ use utils::settings::Settings;
 pub trait DataStorage: Send + Sync {
     async fn get_data(&self, key: &str) -> Result<Bytes>;
     async fn put_data(&self, data: Bytes, key: &str) -> Result<()>;
-    async fn build_test_bucket(&self, bucket_name: &str) -> Result<()>;
+    async fn build_bucket(&self, bucket_name: &str) -> Result<()>;
+    async fn setup(&self, bucket_name: &str) -> Result<()> {
```
apoorvsadana, line 25: the bucket name won't be a param to setup no? the setup function should know what buckets to create in my understanding

crates/orchestrator/src/queue/sqs/mod.rs
```diff
@@ -54,6 +63,166 @@ impl QueueProvider for SqsQueue {
...
+}
+
+const VISIBILITY_TIMEOUT: i32 = 1800; // 30 minutes in seconds
+const DLQ_NAME: &str = "madara_orchestrator_job_handle_failure_queue";
+const MAX_RECEIVE_COUNT: &str = "5";
+
+impl SqsQueue {
+    async fn get_aws_configs() -> SdkConfig {
+        let region_provider = RegionProviderChain::default_provider();
+        from_env().region(region_provider).load().await
```
apoorvsadana, line 104: from what i remember, we pass down AWS config normally from `config.rs`, can we not do something similar here?

crates/orchestrator/src/alerts/aws_sns/mod.rs
```diff
@@ -32,4 +33,13 @@ impl Alerts for AWSSNS {
         self.client.publish().topic_arn(self.topic_arn.clone()).message(message_body).send().await?;
         Ok(())
     }
+
+    async fn setup_alerts(&self) -> color_eyre::Result<()> {
+        let settings_provider = EnvSettingsProvider {};
+        let response =
+            self.client.create_topic().name(settings_provider.get_settings_or_panic("SNS_TOPIC_NAME")).send().await?;
+        let topic_arn = response.topic_arn().unwrap_or_default();
+        log::debug!("SNS topic created. Topic ARN: {}", topic_arn);
```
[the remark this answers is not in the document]
heemankv: use `tracing`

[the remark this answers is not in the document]
ocdbytes: tracing is not needed because this is a setup function.

crates/orchestrator/src/data_storage/mod.rs
ocdbytes: It is better to have it taken as a param in case we wanted to create a new bucket.

crates/orchestrator/src/queue/sqs/mod.rs
```diff
@@ -54,6 +63,166 @@ impl QueueProvider for SqsQueue {
...
+
+    async fn create_event_bridge_client() -> aws_sdk_eventbridge::Client {
+        let region_provider = Region::new(get_env_var_or_panic("AWS_REGION"));
+        let creds = EnvironmentVariableCredentialsProvider::new().provide_credentials().await.unwrap();
+        let config = from_env().region(region_provider).credentials_provider(creds).load().await;
+        aws_sdk_eventbridge::Client::new(&config)
+    }
+
+    fn get_queue_configs() -> Vec<QueueConfig> {
+        vec![
```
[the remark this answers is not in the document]
ocdbytes: But in current design we only have one dlq right ?

crates/orchestrator/src/queue/sqs/mod.rs
```diff
@@ -54,6 +63,166 @@ impl QueueProvider for SqsQueue {
...
+                name: "madara_orchestrator_update_state_job_verification_queue".to_string(),
+                needs_dlq: true,
+            },
+            QueueConfig { name: "madara_orchestrator_worker_trigger_queue".to_string(), needs_dlq: true },
+        ]
+    }
+
+    async fn create_queue(client: &Client, queue_name: &str, redrive_policy: Option<&str>) -> Result<String> {
+        let mut attributes = HashMap::new();
+        attributes.insert(QueueAttributeName::VisibilityTimeout, VISIBILITY_TIMEOUT.to_string());
```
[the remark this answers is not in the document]
ocdbytes: ok but do we take it as an env ? or should I define it as an constant ?

crates/orchestrator/src/setup/mod.rs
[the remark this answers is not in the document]
ocdbytes: but this is a different command right it is a different environment. I did everything in setup module such that It will be only run once.

crates/orchestrator/src/setup/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+            let config = setup_config(ConfigType::AWS).await;
+            let sqs = Box::new(crate::queue::sqs::SqsQueue {});
+            sqs.setup(config, QUEUE_VISIBILITY_TIMEOUT, QUEUE_MAX_RECEIVE_COUNT).await?
+        }
+        _ => panic!("Unsupported Queue Client"),
+    }
+    log::info!("Queues setup completed ✅");
+
+    log::info!("Setting up event bridge");
+    match get_env_var_or_panic("EVENT_BRIDGE_PROVIDER").as_str() {
```
apoorvsadana, line 73: event bridge is AWS specific, cron is a better word

crates/orchestrator/src/setup/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+        "event_bridge" => {
```
apoorvsadana, line 74: move to a cron provider as it will be better

crates/orchestrator/src/setup/worker_triggers/event_bridge.rs
```diff
@@ -0,0 +1,62 @@
...
+    worker_trigger_queue_name: &str,
+) -> color_eyre::Result<()> {
+    let config = match config {
+        SetupConfig::AWS(config) => config,
+        _ => panic!("Unsupported Event Bridge configuration"),
+    };
+    let event_bridge_client = aws_sdk_eventbridge::Client::new(config);
+    let sqs_client = aws_sdk_sqs::Client::new(config);
+
+    event_bridge_client
```
apoorvsadana, line 21: create_cron(rule_name , cron_time: Duration, target_queue_name: String, message: String)

- `rate(1 minute)` should be created inside the event bridge implementation
- queue arn can be fetched inside create_cron
- message string will be built in setup and sent to create cron
- target_queue_name will be sent by setup to create cron
- cron_time will be hardcoded in setup and sent to create_cron

crates/orchestrator/src/cron/event_bridge.rs
```diff
@@ -0,0 +1,102 @@
...
+        let config = match config {
+            SetupConfig::AWS(config) => config,
+            _ => panic!("Unsupported Event Bridge configuration"),
+        };
+        let event_bridge_client = aws_sdk_eventbridge::Client::new(config);
+        let sqs_client = aws_sdk_sqs::Client::new(config);
+
+        event_bridge_client
+            .put_rule()
+            .name(WORKER_TRIGGER_RULE_NAME)
```
apoorvsadana, line 35: should not hardcode work trigger name herer

crates/orchestrator/src/cron/event_bridge.rs
```diff
@@ -0,0 +1,102 @@
...
+        // Create the EventBridge target with the input transformer
+        let input_transformer =
+            InputTransformer::builder().input_paths_map("$.time", "time").input_template(message).build()?;
+
+        event_bridge_client
+            .put_targets()
+            .rule(WORKER_TRIGGER_RULE_NAME)
+            .targets(
+                Target::builder()
+                    .id(format!("worker-trigger-target-{:?}", worker_trigger_type))
```
apoorvsadana, line 59: do we need a `worker_trigger_type`?

crates/orchestrator/src/cron/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+use lazy_static::lazy_static;
+
+use crate::queue::job_queue::{WorkerTriggerMessage, WorkerTriggerType};
+use crate::setup::SetupConfig;
+
+pub mod event_bridge;
+
+lazy_static! {
+    pub static ref CRON_DURATION: Duration = Duration::from_mins(1);
+    pub static ref TARGET_QUEUE_NAME: String = String::from("madara_orchestrator_worker_trigger_queue");
```
apoorvsadana, line 13: can we use the same constant here and the queue file
also add TODO as you mentioned here that we can take this from clap later on

crates/orchestrator/src/queue/mod.rs
```diff
@@ -6,11 +6,87 @@ use std::time::Duration;
...
 use crate::config::Config;
 use crate::jobs::JobError;
+use crate::setup::SetupConfig;
+
+#[derive(Clone)]
+pub struct QueueConfig {
+    pub name: String,
+    pub visibility_timeout: i32,
+    pub max_receive_count: Option<i32>,
+    pub dlq_name: Option<String>,
```
apoorvsadana, line 22: change to DlqConfig

apoorvsadana, line 22: is it better to use &str?

apoorvsadana:
change expects to throwing errors

crates/orchestrator/src/cron/event_bridge.rs
ocdbytes: Ok so when we need to remove this trigger in future or if we want to reset the setup we would need this id. This particular type of format is easier to construct.

crates/orchestrator/src/cron/mod.rs
ocdbytes: we don't have identical constant with `TARGET_QUEUE_NAME` but I have added the comment here.

crates/orchestrator/src/queue/mod.rs
ocdbytes: changed to &str

ocdbytes: If I use string then we need to specify lifetime here. Ig we can move forward with String.

apoorvsadana (approved):

crates/orchestrator/Cargo.toml
```diff
@@ -20,6 +20,7 @@ aws-config = { workspace = true, features = ["behavior-version-latest"] }
 aws-credential-types = { version = "1.2.1", features = [
   "hardcoded-credentials",
 ] }
+aws-sdk-eventbridge.workspace = true
```
heemankv, line 23: let's use same format everywhere either 
.workspace or = { workspace = true, }

crates/orchestrator/src/alerts/aws_sns/mod.rs
```diff
@@ -32,4 +32,11 @@ impl Alerts for AWSSNS {
         self.client.publish().topic_arn(self.topic_arn.clone()).message(message_body).send().await?;
         Ok(())
     }
+
+    async fn create_alert(&self, topic_name: &str) -> color_eyre::Result<()> {
+        let response = self.client.create_topic().name(topic_name).send().await?;
+        let topic_arn = response.topic_arn().expect("Topic Not found");
+        log::info!("SNS topic created. Topic ARN: {}", topic_arn);
```
heemankv, line 39: use tracing

Mohiiit (approved):

PR #193, opened 2024-12-18
Readme Updated

Mohiiit:
Readme updated with the steps to setup and run tests locally


README.md
```diff
@@ -4,11 +4,42 @@ The Madara orchestrator is designed to be an additional service which runs in
 parallel to Madara and handles various critical jobs that ensure proper block
 processing, proof generation, data submission and state transitions.
 
+> 📝 **Note**: These instructions are verified for Ubuntu systems. While most steps remain similar
```
apoorvsadana, line 7: should we say Ubuntu AMD?

README.md
```diff
@@ -73,26 +104,150 @@ The system uses dedicated queues for managing different job phases:
...
+
+> 🚨 **Important Note**: Currently, Madara doesn't support the `get_storage_proof` endpoint.
+>
+> Therefore, you need to run Pathfinder alongside Madara:
+>
+> - Madara will run in sequencer mode
+> - Pathfinder will sync with Madara
+> - The orchestrator will use Pathfinder's RPC URL for SNOS and state update fetching
+>
+> This setup ensures proper functionality until the `get_storage_proof` endpoint is implemented in Madara.
```
apoorvsadana, line 140: to be precise, till SNOS is adapted to work with `get_storage_proof` RPC endpoint

README.md
```diff
@@ -73,26 +104,150 @@ The system uses dedicated queues for managing different job phases:
...
 
-2. **Prover Service**
+1. **Install System Dependencies**
 
-   - ATLANTIC must be running
+   ```bash
+   # Ubuntu/Debian
+   sudo apt-get update
+   sudo apt install build-essential openssl pkg-config libssl-dev
+   sudo apt install python3.9 python3.9-venv python3.9-distutils libgmp-dev python3.9-dev
```
apoorvsadana, line 152: maybe a comment that python3.9 is needed for building SNOS?

README.md
```diff
@@ -73,26 +104,150 @@ The system uses dedicated queues for managing different job phases:
...
+   docker run -d -p 27017:27017 mongo
+
+   # Start Localstack
+   docker run -d -p 4566:4566 localstack/localstack@sha256:763947722c6c8d33d5fbf7e8d52b4bddec5be35274a0998fdc6176d733375314
+
+   # Start Anvil in a separate terminal
+   anvil --block-time 1
+   ```
+
+4. **Setup Mock Proving Service**
```
apoorvsadana, line 213: oh nice! so we've tested with this and we can run it e2e (not the test case, actually run it?)

README.md
```diff
@@ -73,26 +104,150 @@ The system uses dedicated queues for managing different job phases:
...
+
+   ```bash
+   # Create a simple HTTP server that always returns 200
+   # Example using Python
+   python3 -m http.server 8888 &
```
apoorvsadana, line 218: oh interesting, so we don't need that mock prover service arun wrote?

README.md
```diff
@@ -73,26 +104,150 @@ The system uses dedicated queues for managing different job phases:
...
+
+   a. **Using Test Implementation**
+
+   - Reference the implementation in `e2e-tests/tests.rs` which contains `AnvilSetup`
+   - This handles the deployment of both the StarkNet core contract and verifier contract
+
+   ```rust
+   // From e2e-tests/tests.rs
+   let anvil_setup = AnvilSetup::new();
+   let (starknet_core_contract_address, verifier_contract_address) = anvil_setup.deploy_contracts().await;
```
apoorvsadana, line 236: the e2e test case now registers only a specific fact hash, not something that always returns true, should add something in `scripts` to deploy this dummy verifier?

README.md
```diff
@@ -73,26 +104,150 @@ The system uses dedicated queues for managing different job phases:
...
+
+   - Alternatively, you can deploy the mock verifier directly using Foundry
+   - The contract should implement the same interface as used in the tests
+   - Set the deployed address in your environment:
+
+   ```bash
+   MADARA_ORCHESTRATOR_VERIFIER_ADDRESS=<deployed-contract-address>
+   ```
+
+Note: The mock services are intended for development and testing purposes only.
```
apoorvsadana, line 249: can add a 🚧 emoji :)

README.md
```diff
@@ -73,26 +104,150 @@ The system uses dedicated queues for managing different job phases:
...
+In production, you'll need to use actual proving services and verifier contracts.
 
 ### Setup Mode
```
apoorvsadana, line 252: do we also want to add setting up pathfinder over here?

heemankv (approved):
LGTM

README.md
heemankv: It's added generally, in [here](https://github.com/madara-alliance/madara-orchestrator/tree/feat/readme?tab=readme-ov-file#system-dependencies)

README.md
Mohiiit: done

README.md
Mohiiit: done

README.md
Mohiiit: updated

README.md
Mohiiit: it was not working, changed this

README.md
Mohiiit: done

README.md
Mohiiit: added

README.md
Mohiiit: added

README.md
apoorvsadana: That only mentions installing python right? I was saying let's mention why 3.9 because it's an older version (because SNOS only works with that)

README.md
```diff
@@ -73,34 +104,210 @@ The system uses dedicated queues for managing different job phases:
...
+   anvil --block-time 1
+   ```
+
+4. **Setup Mock Proving Service**
+
+   🚧 This setup is for development purposes only:
+
+   ```bash
+   # Start the mock prover service using Docker
+   docker run -d -p 6000:6000 ocdbytes/mock-prover:latest
```
apoorvsadana, line 223: let's use a karnot/madara-alliance naemspace

crates/orchestrator/src/cli/mod.rs
```diff
@@ -403,8 +403,19 @@ pub mod validate_params {
...
+        if (aws_event_bridge_args.aws_event_bridge_rule || aws_event_bridge_args.aws_event_bridge_schedule)
+            && aws_config_args.aws
+        {
+            let cron_type = if aws_event_bridge_args.aws_event_bridge_rule {
+                EventBridgeType::Rule
+            } else if aws_event_bridge_args.aws_event_bridge_schedule {
+                EventBridgeType::Schedule
+            } else {
+                panic!("No Cron service selected!")
+            };
```
apoorvsadana, line 415: why not make a clap arg called event_bridge_type and make that an enum

Mohiiit: we would still need a boolean for event_bridge, just like we have to all other services cli args

we can change it to option<string> with additional boolean which makes sure that if the boolean is true all other args related to event bridge needs to be present

README.md
Mohiiit: as discussed, deferring as of now

README.md
Mohiiit: done

apoorvsadana (approved):

PR #194, opened 2024-12-18
Remove madara and pathfinder submodules

notlesh:
This PR removes the `madara` and `pathfinder` git submodules.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## Added
 
+- Removed unused git submodules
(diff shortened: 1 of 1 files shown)


apoorvsadana:
@notlesh do we want to remove this right now or when we're moving everything to a monorepo? Removing this right now will also need us to modify the `Makefiel`

notlesh:
> @notlesh do we want to remove this right now or when we're moving everything to a monorepo? Removing this right now will also need us to modify the `Makefile`

I think it's fine to wait. I opened this mostly to see if CI would fail (and that would help me understand why the submodules existed in the first place.)

What would need to change about the `Makefile`? At first glance, it looks like it uses Madara through Docker.

PR #197, opened 2024-12-24
Expect removed from various places

Mohiiit:
Removed expect and added wrap for the error

### DIFF
--- a/crates/orchestrator/src/data_storage/aws_s3/mod.rs
+++ b/crates/orchestrator/src/data_storage/aws_s3/mod.rs
@@ -6,6 +6,7 @@ use aws_sdk_s3::primitives::ByteStream;
 use aws_sdk_s3::types::{BucketLocationConstraint, CreateBucketConfiguration};
 use aws_sdk_s3::Client;
 use bytes::Bytes;
+use color_eyre::eyre::Context;
 use color_eyre::Result;
 
 use crate::data_storage::DataStorage;
@@ -50,8 +51,20 @@ impl AWSS3 {
 impl DataStorage for AWSS3 {
     /// Function to get the data from S3 bucket by Key.
     async fn get_data(&self, key: &str) -> Result<Bytes> {
-        let response = self.client.get_object().bucket(&self.bucket).key(key).send().await?;
-        let data_stream = response.body.collect().await.expect("Failed to convert body into AggregatedBytes.");
+        let response = self
+            .client
+            .get_object()
+            .bucket(&self.bucket)
+            .key(key)
+            .send()
+            .await
+            .context(format!("Failed to get object from bucket: {}, key: {}", self.bucket, key))?;
+
+        let data_stream = response.body.collect().await.context(format!(
+            "Failed to collect body into AggregatedBytes for bucket: {}, key: {}",
+            self.bucket, key
+        ))?;
+
         tracing::debug!("DataStorage: Collected response body into data stream from {}, key={}", self.bucket, key);
         let data_bytes = data_stream.into_bytes();
         tracing::debug!(
@@ -74,7 +87,8 @@ impl DataStorage for AWSS3 {
             .body(ByteStream::from(data))
             .content_type("application/json")
             .send()
-            .await?;
+            .await
+            .context(format!("Failed to put object in bucket: {}, key: {}", self.bucket, key))?;
 
         tracing::debug!(
             log_type = "DataStorage",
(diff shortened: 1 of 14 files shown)


ocdbytes (approved):
lgtm!!

crates/orchestrator/src/data_storage/aws_s3/mod.rs
```diff
@@ -88,7 +102,12 @@ impl DataStorage for AWSS3 {
 
     async fn create_bucket(&self, bucket_name: &str) -> Result<()> {
         if self.bucket_location_constraint.as_str() == "us-east-1" {
-            self.client.create_bucket().bucket(bucket_name).send().await?;
+            self.client
+                .create_bucket()
+                .bucket(bucket_name)
+                .send()
+                .await
+                .context(format!("Failed to create bucket: {} in us-east-1", bucket_name))?;
```
heemankv, line 110: Can we try to change this 
```
 if self.bucket_location_constraint.as_str() == "us-east-1" {
```
and compare the enums provided by AWS.

Accordingly change the `context` to not hardcode `us-east-1`

Note:  I tried doing this in a recent PR but the code was failing after
maybe you can try and this time it might work ?
WDYT @Mohiiit ?

crates/orchestrator/src/queue/job_queue.rs
```diff
@@ -231,7 +231,8 @@ fn parse_worker_message(message: &Delivery) -> Result<Option<WorkerTriggerMessag
         .borrow_payload()
         .ok_or_else(|| ConsumptionError::Other(OtherError::from("Empty payload".to_string())))?;
     let message_string = String::from_utf8_lossy(payload).to_string().trim_matches('\"').to_string();
-    let trigger_type = WorkerTriggerType::from_str(message_string.as_str()).expect("trigger type unwrapping failed");
+    let trigger_type = WorkerTriggerType::from_str(message_string.as_str())
+        .map_err(|e| ConsumptionError::Other(OtherError::from(e)))?;
```
heemankv, line 235: Let's add a `wrap_err` here as well,
seems like a good practice to follow 
WDYT @Mohiiit  ?

crates/prover-clients/atlantic-service/src/client.rs
```diff
@@ -85,8 +85,7 @@ impl AtlanticClient {
             )
             .send()
             .await
-            .map_err(AtlanticError::AddJobFailure)
-            .expect("Failed to add job");
+            .map_err(AtlanticError::AddJobFailure)?;
```
heemankv, line 88: Same, can add `wrap_err` here

WDYT on adding `wrap_err` throughout the code @Mohiiit ?

crates/orchestrator/src/queue/job_queue.rs
Mohiiit: valid added

crates/prover-clients/atlantic-service/src/client.rs
Mohiiit: I don't think it makes sense here, because wrap_err is used to provide extra context to the error, in here the only context is that the add job failure happened which is already handled by the AtlanticError here

apoorvsadana (approved):

crates/orchestrator/src/data_storage/aws_s3/mod.rs
Mohiiit: out of scope for the current PR, creating issue for this one

crates/prover-clients/atlantic-service/src/client.rs
Mohiiit: > wrap_err is used to provide extra context to the error

this still stands but I changed the http-client implementation a bit to provide better error messages, lmk wdyt? @heemankv

crates/orchestrator/src/data_storage/aws_s3/mod.rs
heemankv: Please link the issue here.

crates/prover-clients/atlantic-service/src/client.rs
heemankv: works for me

apoorvsadana (approved):

crates/orchestrator/src/data_storage/aws_s3/mod.rs
Mohiiit: https://github.com/madara-alliance/madara-orchestrator/issues/201

heemankv (approved):
LGTM

PR #203, opened 2025-01-07
fixed: instrumentation

heemankv:
The previous fix did not work for `process_job` because `job` that was being sent was not updated with the `external_id` received, we could fetch it again, but doing a DB operation for metrics is a heavy call, hence here we are sending individual components with their most updated values.


ocdbytes (approved):
lgtm !!!

